PR #2, opened 2017-03-11
Removed dependency from Google Chrome

### LINKED ISSUE #1: Use built-in Python webbrowser instead of os.system
Python 2 and 3 have a built-in library named [`webbrowser`](https://docs.python.org/3.6/library/webbrowser.html) which would allow people to open links in their default browser, rather than forcing Google Chrome by using `os.system`.

jinankjain:
Fixes #1

### DIFF
--- a/mapps.py
+++ b/mapps.py
@@ -1,6 +1,7 @@
 import os, json, requests
 from colorama import init
 from colorama import Fore, Back, Style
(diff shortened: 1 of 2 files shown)


sukeesh:
I have corrected this in latest pull request. Hence, closing this.

PR #32, opened 2017-03-19
Osx fix issue 3

bwh91:
I have added a variable to track what platform the application is running on. I am specifically checking if it is running "darwin" if that is the case it's running on OS X. Where platform specific commands need to be run I added a if/else block to check which platforms running. Specifically "Open Camera", "Increase Volume", "Decrease Volume", "Hotspot Start", "Hotspot Stop" and "Check Ram". All functions behave the same for either linux or mac with the exception of the "Hotspot" functions. On OS X automatically starting and stopping a hotspot is blocked by the OS. The work around for this is to open directly the system dialog to set this manually. I have validated on OS X 10.12 and on Ubuntu 16.4. There are no extra dependancies added with these changes.


StefanosChaliasos:
Does it still work on OS X after all the new feature that we had added to Jarvis?

PR #45, opened 2017-03-24
Revert "Expand capabilities of ToDo list"

StefanosChaliasos:
Reverts sukeesh/Jarvis#34


5n8ke:
The error is due to format changes in todolist.txt.
Previously it was filled with custom formatting. Now it uses JSON. I will add a fix so that an existing file without json will be backed up and cleared.
I don't think implementing a converter from the old format to json will be feasable.

PR #47, opened 2017-03-24
Ai interaction

bwh91:
I have added a command called "asked jarvis" it opens up a new loop that the user can interact with jarvis as an AI. It is pretty simple now but I wanted some input on this before I went further. It is fully functional but the aiml file needs to be better. If its pushed to the master branch I will add to it to make it better. This functionality could also be used to make jarvis more human like in the interaction instead of always hard coding responses.


StefanosChaliasos:
I merge it so anyone can contribute to it but please keep improving it.

PR #56, opened 2017-04-06
Adding tests

bwh91:
I have added the tests directory with tests for memory and for the aiml packages. I also updated the readme to explain how to run the tests. Using the two tests I have added as a template it should be pretty straight forward adding more tests.


StefanosChaliasos:
Great work. Can you please fix the tests so there aren't any failure?
screenshot from 2017-04-06 13-19-55

bwh91:
The 'TEST' response was added to the new the new default AIML file that I forgot to add to the PR.... Sorry about that... I will fix that issue tonight...

PR #59, opened 2017-04-08
Test fix

bwh91:
I fixed the test issues you had found with the aiml files. I also added two functions to brain.py to create and remove brain files.


StefanosChaliasos:
Please leave a space after # on inline comments (https://www.python.org/dev/peps/pep-0008/#inline-comments). I fixed it in that PR.

PR #71, opened 2017-05-08
Added the umbrella function

LuisFSilva:
**When asked, Jarvis will inform us if we need an umbrella to go outside or not.**

jarvisumbrella

Please review my code as I'm not the most experienced.

Also, please check if the output text of the function (including help_umbrella) is good enough.


StefanosChaliasos:
Great work!! Please add your name to contributors.md [1].

[1] https://github.com/sukeesh/Jarvis/blob/master/contributors.md

PR #77, opened 2017-05-09
Added 'weather in' as part of the 'check' function

LuisFSilva:
jarvisweatherin


StefanosChaliasos:
It seems like it doesn't work properly. 
screenshot from 2017-05-10 01-09-37

StefanosChaliasos:
My fault!! We need to put check in front of weather! Great Job @LuisFSilva . We might see how we could make it work without check in the future

LuisFSilva:
I felt like it was a good way to integrate the underused check function. If you want I can change it.

Btw, I already did the same but with time (check time in). You can see it in one of my branches. I'll PR it later today. If you want I can change both of them to not use the 'check' command

StefanosChaliasos:
For me it's ok with the check command. @sukeesh  any thoughts?

sukeesh:
@StefanosChaliasos @LuisFSilva  
without check will be good, I guess.

Issue #73, opened 2017-05-09
Convert the weather to Fahrenheit if the location is in US


StefanosChaliasos:

LuisFSilva:
I started working on this today and I got something the works with the normal function of weather (take a look in one of my [branches](https://github.com/LuisFSilva/Jarvis/tree/Enhancement/Weather)), but have no idea how to do this with the new function 'check time in'. Any suggestions?

LuisFSilva:
Nevermind, got it. Check my PR

StefanosChaliasos:
#81

PR #81, opened 2017-05-12
Weather now shows in Fahrenheit if current location is in US (#73)

LuisFSilva:


StefanosChaliasos:
Great work!! Thank you!!

PR #86, opened 2017-05-27
Distro specific update commands

SShehriyar:
Added option for user to select his/her distribution and update using that distribution's update commands.


SShehriyar:
Found a better way to automatically find distribution name and update.

PR #87, opened 2017-05-27
automatically get distribution name and update using distribution's package manager

SShehriyar:


StefanosChaliasos:
Thank you and sorry for the delay!!

Issue #75, opened 2017-05-09
Create tests for clock method 


StefanosChaliasos:

Josiah-DuBose:
Hello, I'd like to help with this. can you provide more details?

Josiah-DuBose:
For instance, is the clock method somewhere in packages?

StefanosChaliasos:
Yeah my mistake. I mean the check time action (method do_check line 97 Cmdinterpeter.py). You could check if the s is equal to your location then is the time given from the call of timeIn.main(self, s) the same with ctime()?

sladyn98:
I'm new to GitHub .can you tell me how can I be helpful with this issue

StefanosChaliasos:
@sladyn98  The steps you need to follow to contribute to a GitHub project are [here](http://stackoverflow.com/questions/4384776/how-do-i-contribute-to-others-code-in-github). You could try to solve the issue #74 which is easier. You can find more information about unit testing on python in that [documentation](https://docs.python.org/2/library/unittest.html) .

trp07:
Hello,
I believe I completed the required test and am about to initiate a PR.  

Please let me know if I did it incorrectly.  This would be my first PR.  I'm new to coding and am trying to get involved!  

Thanks for posting the request!
-Tim

salman-bhai:
Hey, Is this Issue still up for grabs or is it resolved?

trp07:
I submitted pull request #88 in response to this.

salman-bhai:
Oh okay so it hasn't been merged! Thanks :smile:

PR #88, opened 2017-05-30
Tests timeIn.main against ctime for current locale

trp07:
This is my shot at answering issue #75.

I'm new to coding, and this is my first PR.  Please let me know where I went wrong, so I can learn.  Thank you for the opportunity!

### DIFF
--- a/Jarvis/tests/test_timeIn.py
+++ b/Jarvis/tests/test_timeIn.py
@@ -16,7 +16,6 @@ class TimeInTest(unittest.TestCase):
         timeIn.main(self, "time in Greenwich, UK")
         sys.stdout.seek(0)
         output = sys.stdout.read()
-        result = output.split(' ')[10][0:8]
 
         # What is ctime in you current local?
         time_here = time.ctime(time.time()).split(' ')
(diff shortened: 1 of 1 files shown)


trp07:
I adjusted the test_timIn.py file to run in UTC since the test will probably fail depending on what timezone you run the test in.

StefanosChaliasos:
Can you resolve that please? 
screenshot from 2017-06-02 03-47-52

trp07:
Commit a235c6d99... fixes the index error.  Sorry about that.  

Also to note... Some tests fail due to a possible network delay.  The response coming back from the TimeZoneDB API is sometimes delayed by a second or two, which causes the test to fail against the host's ctime output.  Most tests pass though.  

If this is a problem, any thoughts on how to fix it?

Thank you!

trp07:
I just added one more commit.  I think the reason your test failed in the first place is because I ran the test on May 30, and you ran it on Jun 1.  The difference in time.ctime(time.time()) outputs is that on Jun 1, an extra space is added to the output to make sure the days of the month align at the same position in the string.  This commit now account for whether the day of the month is one or two digits long.

StefanosChaliasos:
Nice work!! Please add your name to [contibutors.md](https://github.com/sukeesh/Jarvis/blob/master/contributors.md) and I will merge your PR.

trp07:
Thank you!  I added myself to contributors.md.

Issue #92, opened 2017-06-02
No error message displayed for two word commands missing 2nd word


anaisabel7:
For two (or more) word commands, such as `disable sound` or `decrease volume`, no error or warning message is displayed if only the first word is found in the command passed to Jarvis, such as:
`decrease`
`please, can you decrease`
`decrease my cat, thank you`

nashid:
is this issue still up for grab?

StefanosChaliasos:
No a solution had been merged

PR #93, opened 2017-06-02
[HELP WANTED] Two word cmds warnings

anaisabel7:
Calling "default" as error message function when the first word of a several word command can be found in the prompt to Jarvis, but not the rest of required words. See issue #92

### DIFF
--- a/Jarvis/tests/test_default.py
+++ b/Jarvis/tests/test_default.py
@@ -9,7 +9,7 @@ class DefaultTest(unittest.TestCase):
         self.jarvis = Jarvis()
         self.action = {"enable": ("sound",)}
 
-    # check that default is called if the value of the word dictionary is not found in
+    # check that default is called if the second word of a two word command is not found in
     # the words_remaining ("enable cat and dog" instead of "enable your sound please")
     def test_call_default_for_wrong_dict_cmd(self):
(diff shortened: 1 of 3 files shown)


Jarvis/tests/test_default.py
```diff
@@ -0,0 +1,44 @@
+import unittest
+from Jarvis import Jarvis
+from mock import patch
+
+
+class DefaultTest(unittest.TestCase):
+
+    def setUp(self):
+        self.jarvis = Jarvis()
```
anaisabel7: I could use some help understanding why this is a problem for **Codacy** given it uses the same syntax as as the test_parser.
The tests pass correctly too.
I cannot figure out how to solve the problem and make Codacy happy.

StefanosChaliasos: It's ok don't worry.

StefanosChaliasos:
can you provide an example of your functionality

anaisabel7:
This is the behaviour previous to my changes (in the master branch):
in_master_branch
And this is how Jarvis behaves in this feature branch:
in_feature_branch

anaisabel7:
@StefanosChaliasos is there anything else you want me to add / change / explain in here? :smiley_cat:

StefanosChaliasos:
Thanks for your contributions

Issue #98, opened 2017-06-05
Music not downloading


anaisabel7:
When trying to use the music feature with a command such as
`music i was made for loving you`
after picking one of the offered choices and confirming the download
`Pick one: 8`
`Download "KISS - I Was Made For Loving You"? (y/n) y`
an error occurred with the following information:
`ERROR: Signature extraction failed: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl/extractor/youtube.py", line 1070, in _decrypt_signature
    video_id, player_url, s
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl/extractor/youtube.py", line 958, in _extract_signature_function
    raise ExtractorError('Cannot identify player %r' % player_url)
ExtractorError: Cannot identify player u'https://www.youtube.com/yts/jsbin/player-vflZ_L_3c/en_US/base.js'; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
 (caused by ExtractorError(u"Cannot identify player u'https://www.youtube.com/yts/jsbin/player-vflZ_L_3c/en_US/base.js'; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.",)); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
`

PR #99, opened 2017-06-05
Update version of youtube-dl

anaisabel7:
An update on the version of youtube-dl was required for the music feature to work.
This should solve issue #98 

Also, a problem was found in the correct identification of the song name, which has now been sorted. In this respect, this is the behaviour in master:
song_master
and this is the behaviour in this feature branch now:
song_feature

### DIFF
--- a/Jarvis/packages/music.py
+++ b/Jarvis/packages/music.py
@@ -5,11 +5,11 @@ from colorama import Fore
 
 
 def play(data):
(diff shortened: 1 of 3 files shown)


anaisabel7:
Regarding the **Codacy** warning, I am pretty sure the patch has to go on "os" (shell).
Please, let me know if there is anything to change regarding that warning.

Issue #100, opened 2017-06-07
Jarvis breaks when using the near command


anaisabel7:
When trying to run the command `restaurants near me` within Jarvis, Jarvis breaks. This is the last part of the traceback:
```
  File "Jarvis/Jarvis.py", line 143, in executor
    self.cmdloop(self.first_reaction_text)
  File "/usr/lib/python2.7/cmd.py", line 142, in cmdloop
    stop = self.onecmd(line)
  File "/usr/lib/python2.7/cmd.py", line 221, in onecmd
    return func(arg)
  File "Jarvis/CmdInterpreter.py", line 319, in do_near
    near_me.main(data)
  File "Jarvis/packages/near_me.py", line 7, in main
    things = " ".join(wordList[0:wordIndex(data, "near")])
  File "Jarvis/utilities/GeneralUtilities.py", line 7, in wordIndex
    return wordList.index(word)
ValueError: 'near' is not in list
```

anaisabel7:
Thanks to the latest changes (see the PR merged above), this is no longer an issue.

PR #101, opened 2017-06-07
Solving the problems with the near command

anaisabel7:
The general parsing of the command within Jarvis meant, for the near command, the information previous to the 'near' keyword was not passed to the near_me.main function, which requires it.
An example is in a command such as `restaurants near me`, the word `restaurants` was not passed to near_me.main.

In this PR, the information previous to 'near' is passed to near_me.main.
The expected syntax of the data passed to near_me.main was also slightly altered to better match the current processing of the command information within Jarvis (i.e. the keyword --- such as 'near' --- is not passed to the do_xxx functions).

This should solve issue #100

### DIFF
--- a/Jarvis/Jarvis.py
+++ b/Jarvis/Jarvis.py
@@ -17,7 +17,7 @@ from CmdInterpreter import CmdInterpreter
 """
 
 
-class Jarvis(CmdInterpreter):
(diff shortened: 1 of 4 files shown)


anaisabel7:
Behaviour in master:
near_master
Behaviour in this feature branch:
feature_branch_near

Issue #12, opened 2017-03-14
Music downloaded as .webm


abrahammurciano:
Music is downloaded as .webm and cannot be played by vlc, is there any way to get it to download as mp4?

anaisabel7:
In the PR above there is a way to download music in mp3 rather than .webm (after re-doing the setup) but I am not sure if it is the desired way to solve this issue....

PR #105, opened 2017-06-19
Music mp3

anaisabel7:
The command to set up the prerequisites correctly has been changed to proceed with the installation that allows music to be downloaded in mp3.

I am not sure if this is a desired feature, but it is a way to solve issue #12


StefanosChaliasos:
Thank you and sorry for the delay!!

PR #112, opened 2017-06-26
replaced Notify with notify2

alokst:
Unable to use Notify. Replacing it with notify2 fixes the issue. Also, had to init it before the for loop.

### DIFF
--- a/Jarvis/packages/reminder.py
+++ b/Jarvis/packages/reminder.py
@@ -3,9 +3,8 @@
 from datetime import datetime as dt
 from uuid import uuid4
 from threading import Timer
-import gi
-gi.require_version('Notify', '0.7')
-from gi.repository import Notify
+import notify2
+
 
 from fileHandler import writeFile, readFile, str2date
 from utilities.lexicalSimilarity import scoreSentence, compareSentence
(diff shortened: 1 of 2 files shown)


StefanosChaliasos:
Can you please update the requirements.txt?

alokst:
python-notify2 is not distributed using pip. Hence, I did not update requirements.txt.
One has to do **_sudo apt-get install python-notify2_**

appi147:
@StefanosChaliasos  in reminder.py gi module throws attribute error. Using notify2 and reminder.py by @alokst fixed this

StefanosChaliasos:
@alokst  I tried with that pip [module](https://pypi.python.org/pypi/notify2) and worked.

alokst:
Hi @StefanosChaliasos . it doesn't seem to work for me. If I do **pip install python-notify2** it is unable to find the distribution and if I do **pip install notify2** it installs but when I import it, it requires another module called dbus. Anyway, what do you propose?

StefanosChaliasos:
Just add the apt-get command to [setup](https://github.com/sukeesh/Jarvis/blob/master/Setup) file

sukeesh:
@alokst Please resolve conflicts

PR #131, opened 2017-07-01
Added Fedora setup Instructions

### LINKED ISSUE #128: Issue in Setup
As the setup file is only targeted for ubuntu/Debian based user. If it is executed on the Redhat based machine. It is showing error messages, because the package downloader for them is `dnf`. So there is a need to make it consistent across platforms. So that all linux distros can use this project.

yashLadha:
Fixes #128

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/sh
+
+OS=$(lsb_release -si)
+if [[ "$OS" == "Fedora" ]]; then
+  sudo dnf install ffmpeg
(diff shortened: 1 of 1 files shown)


yashLadha:
@sukeesh @StefanosChaliasos Review this!!

yashLadha:
Currently I am on a Fedora machine, so i was able to test only on that . Other systems, install instructions please post in the comments, so that i can integrate them. Windows and Mac are also preferred. 😄

StefanosChaliasos:
That's great!!! thank you!

PR #133, opened 2017-07-02
improve setup.sh

appi147:
- #/bin/bash is the location of bash

- elif for Ubuntu

- else condition for other os


appi147:
fixes #128

PR #152, opened 2017-07-23
fb_login checkout

umangahuja1:
I am new with contributions so if there is something regarding contributions let me know. Have a look at the script, it works fine.
Plus I have updated requirements file with Selenium.


StefanosChaliasos:
how can I do login from Jarvis?

yashLadha:
Remove the .idea files as it is not necessary for the current codebase.

yashLadha:
Implement some encryption and hashing so that password is not easily shown in the file. May be there is a risk of hacking.

umangahuja1:
I will look into it. Can anyone tell about the tests failed and the reason. I am not familiar with these projects yet.

umangahuja1:
@StefanosChaliasos I didn't get what you mean?

StefanosChaliasos:
You can remove the .idea file by add to .gitignore file the .idea extension. What i meant was how we could check that function?

umangahuja1:
Oh I forgot to add it into the driver program. I just checked it locally. I will update it and add here.

jarviscli/CmdInterpreter.py
```diff
@@ -611,3 +613,10 @@ def help_umbrella(self):
         """Print info about umbrella command."""
         print_say(
             "If you're leaving your place, Jarvis will inform you if you might need an umbrella or not.", self, Fore.BLUE)
+
+    def do_fb_login(self,s):
```
appi147: Add `s=None`, in case user, does not give any parameter.

umangahuja1: ok thanks. And can you tell about this failed checks?

appi147:
@umangahuja1 IMO `-` (underscore) in `fb_login` may be a problem. Try to rename the function and push.

appi147:
There was no need to delete `.idea/`
Just add `.idea/` in `.gitignore`

jarviscli/CmdInterpreter.py
```diff
@@ -83,6 +84,7 @@ def __init__(self, first_reaction_text, prompt, first_reaction=True, enable_voic
                         "umbrella",
                         {"update": ("location", "system")},
                         "weather",
+                        "fb_login"
```
appi147: Please add in sorted order

umangahuja1:
**This is the error message mentioning static use**

Used when there is no reference to the class, suggesting that the method could be used as a static function instead If the class method does not reference any of the class attributes it may be more clear to define the method as a static function instead. Attempt using either of the decorators @classmethod or @staticmethod Ex: class Foo(object): ... def bar(self, baz): ... return llama

Try instead to use: class Foo(object): ... @classmethod def bar(cls, baz): ... return llama

And it went over the head. Any help would be great if you know

jarviscli/CmdInterpreter.py
umangahuja1: Ok I will correct it and will keep that in mind in future. And underscore is not the problem I think because there are functions similar to it (having underscore).

appi147:
It's error [R0201](http://pylint-messages.wikidot.com/messages:r0201). I will try to find its solution.

umangahuja1:
Is is due the the fact that module fb_login contains only functions and no classes?

appi147:
@umangahuja1 Is this feature working now?

umangahuja1:
Finally that check is passed.

umangahuja1:
What is the reason for this build failure, I think work is done from my side as far as this PR is concerned. Let me know so that I can submit another PR

StefanosChaliasos:
@umangahuja1 could you update the [NEW_FUNCTIONALITIES.md](https://github.com/sukeesh/Jarvis/blob/master/NEW_FUNCTIONALITIES.md) file with your functionality?

umangahuja1:
Please review and merge

StefanosChaliasos:
screenshot from 2017-07-26 13-44-14

umangahuja1:
sorry I forgot to tell about the driver to be installled, I will include it in requirements'txt.

appi147:
@umangahuja1 driver cannot be added in requirements.txt file. Add `sudo apt-get install chromium-chromedriver` can be added in setup.sh for chromium driver.

umangahuja1:
thanks.

appi147:
your code does not follow PEP8 instructions. Run `coala --ci` from `Jarvis/` to know the errors. Will have to install for coala `pip3 install coala-bears`

umangahuja1:
I have made some mess with the project. Can anyone suggest what to do?

StefanosChaliasos:
You could close that PR and create a new one.

umangahuja1:
I have to fork and commit again by deleting previous repo?

appi147:
@umangahuja1 Please don't work on master when opening a PR. Work on new branch

_Sent from my Xiaomi 2014818 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

umangahuja1:
I will that in mind as this is my first time working on github project. Will send a new PR soon.

umangahuja1:
how to add chromedriver installation for other os.Can anyone tell?

StefanosChaliasos:
#185 duplicate

PR #163, opened 2017-07-24
Resolved tests error

yashLadha:
- Error while importing packages in tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ jarviscli/packages/memory/memory.json
 *.brn
 dist/
 build/
+*.swp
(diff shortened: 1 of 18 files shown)


appi147:
Hi, @yashLadha I see that you are adding `..` before packages import. It all depends on the directory from where you run the program. When `Jarvis` is run, the main program is in `jarviscli/`, all the imports are fine. But if you run from `packages` it will throw `ImportError`. Try to run from `jarviscli` and see if any error is occuring.

yashLadha:
See the directory structure has been changed from past and it has not been implemented in the tests. So I had declared the imports so that it can be accessible however still i am not able to correct test for near_me

appi147:
In last month I was responsible for changes in the directory structure. But when run from `jarviscli` all works fine. near_me was working earlier, now it's failing for me too. Will look into it

yashLadha:
It is not working but the test have not modified according to it.

yashLadha:
I am closing this PR as i need to work more on it.

PR #174, opened 2017-07-27
new feature clear

umangahuja1:
added clear feature to clear jarviscli screen. Review.

### DIFF
--- a/jarviscli/CmdInterpreter.py
+++ b/jarviscli/CmdInterpreter.py
@@ -20,9 +20,9 @@ from packages.memory.memory import Memory
 from packages.shutdown import shutdown_system, cancel_shutdown, reboot_system
 from packages.systemOptions import turn_off_screen, update_system
 from packages.news import News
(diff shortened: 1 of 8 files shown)


jarviscli/CmdInterpreter.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, first_reaction_text, prompt, first_reaction=True, enable_voic
                         "chat",
                         {"check": ("ram", "weather", "time", "forecast")},
                         "chuck",
+                        "clear_scr",
```
appi147: It should have been "clear"

umangahuja1: Is it showing any error or what? I didn't get

appi147: No error, but we have to add "clear" in actions tuple because we have defined "do_clear". It does not matter what you add in tuple, but just for good code to read

PR #203, opened 2017-08-16
fixed bug in check weather

ajf80:
When prompted to enter a city to get weather from, if the city was not found, an error would be thrown in the terminal and jarvis crashed. I changed this so that now a proper message of "city not found" is displayed to the user and jarvis continues to prompt the user for input.


ajf80:
I apologize but I am not sure what exactly is meant by this. I understand that tests failed but the same tests failed in the previous commit as well. Am I wrong?

PR #209, opened 2017-09-06
fix weatherIn test bug

appi147:


appi147:
Build Passing :100: :tada:

Issue #208, opened 2017-09-05
Wiki Search Option


kedarkhetia:
Hi, 
I can see that there is no wiki search option provided in the (BOT). It would be really useful if we have a simple command such as 

`wiki Git`

And it provides simple output which summarize the information and provide a link to wiki page for the search.

Being new to open source, I would like to take this task and contribute to open source.

BradleyJHumble:
https://pypi.python.org/pypi/wikipedia

appi147:
@kedarkhetia Sure create a PR :+1:

kedarkhetia:
@appi147 Can you please assign this to me ? Also, if possible add label 'feature-request' to this issue.

Thanks,
Kedar Khetia.

appi147:
There you go!!

appi147:
How is work going @kedarkhetia

kedarkhetia:
Hi, 

Sorry couldn't work on it as I am busy with office work. Will work on it as soon as I am free.

Thanks,
Kedar Khetia.

Issue #121, opened 2017-06-29
Implement the wikipedia API


StefanosChaliasos:
[documentation](https://pypi.python.org/pypi/wikipedia/)

Prakash2403:
I would like to work on this.

StefanosChaliasos:
Go for it!

kaustav1996:
hi hows the work going?

StefanosChaliasos:
@kaustav1996 you could give it a try!

umangahuja1:
What is the progress?

kaustav1996:
Actually am a little busy with my internship right now , but as soon as I
complete my internship, I will work on this.

AayushGala:
We can use Wikipedia's API to get info instead of using a module

SolaOmi:
Is this feature request still up for grabs?

PR #223, opened 2017-10-09
Implementation of the Wikipedia API

SolaOmi:


appi147:
@SolaOmi Please add `wikipedia` in `requirements.txt`

appi147:
@SolaOmi 

- What is `wiki suggest`, it always says `No article with that name, try another item.`

- Please improve `wiki search` to give results in a proper format, not raw data.

- I don't think various options under `wiki page` are of any use on CLI, such as HTML, Images, Links, etc.

- Summary feature is added twice

- I'll suggest replacing `page` with `content` feature which is provided under `page`

SolaOmi:
Thanks for the feedback @appi147, will get on it asap

SolaOmi:
- Wiki suggest is a function that tries to guess what you are looking for when you misspell something, but I removed it since I don't think it works very well.

- I cut down the wiki command down to the three functions search, summary, and content.
     1. search - returns a list of 10 pages related to a query
     2. summary - returns the summary in plain text from the query's page
     3. content - returns all the plain text on the query's page (excluding images, tables, and other data)

appi147:
Closes #121.
Closes #208

appi147:
@SolaOmi Add this feature in NEW_FUNCTIONALITIES.md and I think this can be merged

SolaOmi:
@appi147 all set!

appi147:
Nice work :+1:

PR #226, opened 2017-10-11
Remove setup files

appi147:
As these files are not needed for general users I think it should be removed from repo, and remain only with sukeesh


appi147:
What do you think @sukeesh @StefanosChaliasos

appi147:
Also, I think contributors.md is of no use as Github has this feature

sukeesh:
Go ahead @appi147

PR #232, opened 2017-10-13
setup.sh: Remove redundant chmod

supergr35:
chmod 755 makes chmod +x redundant, meaning that chmod +x should be removed


appi147:
Thanks for contributing!!

Issue #224, opened 2017-10-10
DBUS error


krishnapaul242:
Error Message while running.
`Traceback (most recent call last):
  File "/home/krishna/anaconda2/lib/python2.7/runpy.py", line 174, in _run_module_as_main "__main__", fname, loader, pkg_name)`
`  File "/home/krishna/anaconda2/lib/python2.7/runpy.py", line 72, in _run_code  exec code in run_globals`
`File "/home/krishna/Desktop/Work/Jarvis/jarviscli/__main__.py", line 2, in <module> import Jarvis`
`File "/home/krishna/Desktop/Work/Jarvis/jarviscli/Jarvis.py", line 5, in <module> from CmdInterpreter import CmdInterpreter`
`File "/home/krishna/Desktop/Work/Jarvis/jarviscli/CmdInterpreter.py", line 17, in <module> from packages.todo import todoHandler`
`File "/home/krishna/Desktop/Work/Jarvis/jarviscli/packages/todo.py", line 5, in <module> from reminder import addReminder, removeReminder`
`File "/home/krishna/Desktop/Work/Jarvis/jarviscli/packages/reminder.py", line 16, in <module> import notify2`
`File "/home/krishna/anaconda2/lib/python2.7/site-packages/notify2.py", line 39, in <module> import dbus`
`ImportError: No module named dbus`

I tried to install some modules but stuck in this dbus module
`dbus` not found in PIP or APT.
`dbus-python` found in PIP but requiring dbus-glib-1, which is not there in PIP.

OS Information -
NAME="Ubuntu"
VERSION="17.04 (Zesty Zapus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 17.04"
VERSION_ID="17.04"
VERSION_CODENAME=zesty
UBUNTU_CODENAME=zesty

Write if you have any query or suggestion regarding this.

appi147:
Steps to be followed to solve : 

- `pip install dbus-python`

- `sudo apt install libdbus-1-dev libdbus-glib-1-dev`
If still not solved comment with error

krishnapaul242:
`Failed building wheel for dbus-python
[...]

PR #237, opened 2017-10-14
Added virtualenv

appi147:
#238

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,13 +4,13 @@ jarvispath=$PWD
 touch jarvis
 chmod +rwx jarvis
 exec 3<> jarvis
-
-    # Let's print some text to fd 3
+    
     echo "#!/bin/bash" >&3
+    echo "source $jarvispath/env/bin/activate" >&3
     echo "python $jarvispath/jarviscli/" >&3
 
-# Close fd 3
 exec 3>&-
+
 sudo mv jarvis /usr/local/bin/jarvis
 
 UNAME=$(uname -s)
@@ -38,8 +38,12 @@ if [[ "$OS" == "Fedora" ]]; then
 elif [[ "$OS" == "Ubuntu" ]] || [[ "$OS" == "LinuxMint" ]]; then
   sudo apt-get install ffmpeg python-imdbpy python-notify2
   sudo apt-get install python-dbus libssl-dev libffi-dev
-  sudo -H pip install -r requirements.txt
   sudo apt-get install chromium-chromedriver
+  sudo pip install virtualenv
+  virtualenv env
+  virtualenv -p /usr/bin/python2.7 env
+  source env/bin/activate
+  pip install -r requirements.txt
 elif [[ "$OS" == "Kali" ]]; then
   apt-get install ffmpeg
   apt-get install python-notify2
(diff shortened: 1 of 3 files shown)


appi147:
@StefanosChaliasos Please review

appi147:
@krishnapaul242 Please verify if this fixes #224

setup.sh
```diff
@@ -38,8 +38,12 @@ if [[ "$OS" == "Fedora" ]]; then
 elif [[ "$OS" == "Ubuntu" ]] || [[ "$OS" == "LinuxMint" ]]; then
   sudo apt-get install ffmpeg python-imdbpy python-notify2
   sudo apt-get install python-dbus libssl-dev libffi-dev
-  sudo -H pip install -r requirements.txt
   sudo apt-get install chromium-chromedriver
+  sudo pip install virtualenv
+  virtualenv env
+  virtualenv -p /usr/bin/python2.7 env
+  source env/bin/activate
+  pip install -r requirements.txt
```
StefanosChaliasos: you forgot sudo here

setup.sh
```diff
@@ -38,8 +38,12 @@ if [[ "$OS" == "Fedora" ]]; then
 elif [[ "$OS" == "Ubuntu" ]] || [[ "$OS" == "LinuxMint" ]]; then
   sudo apt-get install ffmpeg python-imdbpy python-notify2
   sudo apt-get install python-dbus libssl-dev libffi-dev
-  sudo -H pip install -r requirements.txt
   sudo apt-get install chromium-chromedriver
+  sudo pip install virtualenv
+  virtualenv env
+  virtualenv -p /usr/bin/python2.7 env
+  source env/bin/activate
```
StefanosChaliasos: it didn't work for me

setup.sh
appi147: `sudo` in which line?

StefanosChaliasos: 46

appi147: `sudo` will install packages systemwide but we need to install packages in virtual environment

setup.sh
appi147: can you give me errors(screenshot or log)

setup.sh
StefanosChaliasos: Yeah you are right. The problem is that some packages need sudo to install. You can try something like [that](https://askubuntu.com/questions/234758/how-to-use-a-python-virtualenv-with-sudo)

setup.sh
StefanosChaliasos: error_message

appi147: Is it working now?

appi147:
@StefanosChaliasos Is it working now?

setup.sh
appi147: As pointed out in the article, if we need an external script to run as superuser, we have to follow that method, but pip is already created once we create a virtual environment.

StefanosChaliasos:
It worked for me (Ubuntu)

krishnapaul242:
No, It didn't work.

appi147:
I'm on Ubuntu 17. 04 and tested on unity as well as xfce desktop environments

appi147:
@krishnapaul242 if you are on another desktop environment please let us know

PR #239, opened 2017-10-17
Add support for Arch-based linux

eudescar:
Arch and Manjaro


StefanosChaliasos:
Could you use virtualenv to download the requirements?

eudescar:
I can use pip to download dbus-python and imdbpy. There's also a package on pypi called ffmpeg-python but I don't know if it works the same way.

Issue #238, opened 2017-10-14
Add virtualenv


appi147:
- [x] MacOS
- [x] Ubuntu/Mint
- [x] Fedora
- [x] Kali
- [x] Arch

Or add support to other OS

amirkarimi:
Is this about adding virtualenv support in `setup.sh` like [here](https://github.com/sukeesh/Jarvis/blob/master/setup.sh#L20)?

appi147:
yes

umangahuja1:
Is anyone working on it?

appi147:
@umangahuja1 support for unticked OS is still required, Ubuntu and Mac are done

PR #246, opened 2017-10-20
Update setup.sh

eudescar:
- **Check pip install virtualenv root permission**
If the user doesn't have python installed locally, it will be necessary to run `pip install` as root. If the command returns the value **2** (permission error raised), we run the command again with **sudo**.
- **Check OS through `/etc/<package.conf>` and not `lsb_release`**
The `lsb_release -si` command returns the distribution's name. If we rely on that, we would need an `if` condition for each distribution. Besides that, there are some distros that don't have that command out of the box.
Checking the `/etc/<package.conf>` file, we can determine if the distribution is **Arch**, **Debian** or **Fedora** based, thus, we can have a much better support.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ packaging==16.8
 pyowm==2.6.1
 pyparsing==2.2.0
 PySocks==1.6.6
-cryptography==1.8.1
+cryptography>=1.8.1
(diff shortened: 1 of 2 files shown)


appi147:
closes #238

PR #253, opened 2017-10-27
Test in OSX on travis

appi147:


appi147:
Please let dbus-python package be in setup.sh, it messes with travis

PR #269, opened 2017-12-14
Added conditional to hackathon.py

GrantFleming:
Previously when typing the command "hackathon" Jarvis would crash out because it would try to call find_all on upcoming when upcoming was of NoneType. 

I am very much a beginner programmer and was unable to work out why BeautifulSoup seemed unable to populate upcoming but I added a conditional so that if 

upcoming is not None: 
    it does as intended 
else:
   it prints message "No hackathon data found."

therefore it does not crash the program anymore.


appi147:
Nice work!! But one error: `jarviscli/packages/hackathon.py:21:113: E128 continuation line under-indented for visual indent`

GrantFleming:
Hopefully I've fixed that now. Will read up on PEP-8 before submitting anymore contributions. Thanks!

GrantFleming:
It seems there was a network connection issue during the build during testing. Is there any way to re-run the test?

appi147:
Restarted the build!!

PR #283, opened 2017-12-28
There was bug in the cricket module which I have fixed.

sscoder:


appi147:
Why did you remove import of six and codes related to it? Its for python3 compatibility

PR #288, opened 2017-12-28
 cric update v3.5 

sscoder:


appi147:
Everything seems to be working fine. I will test and merge within few days if no bug is found

PR #289, opened 2018-01-13
added 'help' to the list of available actions

### LINKED ISSUE #251: Help Function
When I call the help function by itself it lists all of Jarvis's commands. However,  when I type help ``<command>`` I don't get useful information about the command, and instead it actually evokes the command. Is this just me or a wider issue that needs to be fixed?

aswanipranjal:
Hey @appi147, i think this should close #251. It works for me, I am using MacOSX-13.2. Can you please check if this works?


appi147:
Thanks for contributing!!

PR #291, opened 2018-02-08
Update setup.sh

BradleyWalker:
There was a typo in line 73


appi147:
Thanks for pointing out

Issue #250, opened 2017-10-24
Lyrics - help


umangahuja1:
How is this feature exactly work?

appi147:
@jlvivero Unfortunately this feature is currently not working. Can you fix it?

jlvivero:
I'll check it out, I need to see what broke exactly, unfortunately the feature is dependent on third parties (like who serves the lyrics and the python library).

umangahuja1:
If you need any sort of help let us know here.

jlvivero:
sorry for the delays I had to lend my laptop to my brother because his broke. my linux enviroment is there which is where I test and do development for open source stuff. 

So I updated jarvis and I am having problems even installing the newer version, I had a dbus problem (which I solved by changing and if fi statement in the setup.sh which is a problem btw, and also removing the sudo from pip install dbus-python ).

that finally successfully installed it and when i run jarvis i get the following error:
![screenshot_20171117_093256](https://user-images.githubusercontent.com/7948154/32960596-5aa62d7a-cb7a-11e7-9a86-3ab7eb2193bb.png)

Any ideas? (I wish I had the old version to compare functionality so I could identify easily what broke but unfortunately i had to wipe my laptop after I got it back)

appi147:
imdb package upgraded to python3 only

jlvivero:
Should build again with python3 then?

appi147:
Can you try `pip install IMDbPY==5.1.1` to test if works because it is already cached on my machine

appi147:
@jlvivero

7ayushgupta:
Is this issue currently unassigned? I'd like to work on that issue. 
PS: I am a newbie to the codebase

7ayushgupta:
It's no problem. I'll work a bit on it on my own :smile:

jlvivero:
@7ayushgupta I just got my internet back, how I'll setup my work environment tonight and if you need help or are having trouble doing it, I'll get on it on the weekend

PR #293, opened 2018-02-09
Added macos support for file_organise and help text for lyrics command

CTapas:
- Added support for mac os in file_organise command; file_organise command exited with error "file not found" because it was looking in '/home' dir which is linux only path.
- Added example in help text for lyrics since it was not very clear as to how to use it.


appi147:
closes #250

appi147:
Thanks for contributing

PR #295, opened 2018-02-27
test: added tests for cricket.py to increase code coverage

Sjobom:
refs #9

*These tests make sure most lines are covered by a test in cricket.py
*There could be a lot of improvement in how the results are validated as I found it hard to come up with a good oracle to check against. This is due to the fact that the data given just reflects the latest cricket games and you cannot query a specific game from cricket.py...
* The coverage of cricket.py was 5% before these tests and are 89% now with these in place!


Sjobom:
Sorry, this was supposed to be a PR in our fork...

PR #310, opened 2018-04-14
currencyconv.py

kfragkedaki:
I added a new feature in order to convert a given amount of money from a specific currency of our choice to another currency. For example write:  currencyconv 120 dollars to euro


StefanosChaliasos:
Consider using [forex-python](https://github.com/MicroPyramid/forex-python). It could be used for cryptocurrencies as well.

PR #314, opened 2018-04-29
Featre: Roll a dice, openSUSE and input()

pnhofmann:
Hi,

it's my first pull request - so please tell me if I do something stupid ;).

I added simple command to roll one (or lots of) dice(s).

Since my develop-os is openSUSE I adapted the setup and update system-code.

I think lsb_release -i is a more reliable way to get Distributor id than parsing /etc/lsb-release. lsb_release is required by LSB Specifications - but not DISTRIB_ID in /etc/lsb-release (may be e.g. in /etc/os-release - like on my openSUSE machine).

Finally there is six.moves.input - as Python 2 raw_input and Python 3 input replacement. So no need for ifs for each input ;).

### DIFF
--- a/jarviscli/CmdInterpreter.py
+++ b/jarviscli/CmdInterpreter.py
@@ -2,6 +2,7 @@ from os import system
 from cmd import Cmd
 import signal
 import six
+from six.moves import input
 from platform import system as sys
 from platform import architecture, release, dist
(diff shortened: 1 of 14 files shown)


StefanosChaliasos:
Nice work!! Next time create a different PR for each addition!

PR #315, opened 2018-04-29
Currencyconv.py

kfragkedaki:
New feature and its unittest is created.
This feature is converting a given amount from a currency of our choice to another currency using forex-python library.


jarviscli/packages/currencyconv.py
```diff
@@ -1,42 +1,48 @@
...
+    with open(os.path.join(FILE_PATH, "../../data/currencies.csv"), mode='r') as infile:
+        reader = csv.reader(infile)
+        mydict = {r: row[2] for row in reader for r in row}
+    return mydict
+
+def get_currency(prompt, currencies):
+
+    first_time = 1
+    while True:
+        c = raw_input(prompt).upper()
```
StefanosChaliasos: Use input from six.moves

appi147:
Nice work @kfragkedaki

PR #318, opened 2018-05-19
301 docker

jpigree:
Adds Dockerfile from which you can create a Jarvis Docker image from the currently checked out Jarvis code.
I faced multiple issues mainly because in container distribution have many differences from the host ones (systemd, dbus, ...).

A Makefile is also here to document how to build the Dockerfile and run the image.

As a side note, I refactored a bit the setup.sh to pass more cleanly the Python version used.


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN apt-get update && \
+    apt-get -y install \
+            dbus-x11 \
+            libdbus-1-dev \
+            libespeak1 \
+            libnotify-cil-dev \
+            notification-daemon \
+            sudo \
+            xorg
+
```
gucharbon: Remove apt cache after installing depndencies as mentioned in dockerfile best practices (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) 
```
RUN apt-get update && \
    apt-get -y install \
            dbus-x11 \
            libdbus-1-dev \
            libespeak1 \
            libnotify-cil-dev \
            notification-daemon \
            sudo \
            xorg  && \
     rm -rf /var/lib/apt/lists/*
```

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN pip3 install virtualenv
```
gucharbon: Use pip without cache:

`RUN pip3 install --no-cache-dir virtualenv`

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+
+COPY . /app
+WORKDIR /app
+
+RUN sudo /bin/bash ./docker.sh
```
gucharbon: Why use sudo ? Are you not still root user at this point ? 
In any way, best pratice is to reduce number of docker layers. Lines 20 to 24 could be added in single layer.

`RUN  /bin/bash ./docker.sh && \
    useradd -Um jarvis && \
    echo "jarvis  ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \
    chown -R jarvis /app 
` 

By the way since you add a line into /etc/sudoers you're definetely root at this point, remove previous sudo.

setup.sh
```diff
@@ -1,13 +1,26 @@
...
-case ${answer:0:1} in
-2 )
-    echo "Selected python version 2"
-;;
-* )
-    echo "Selected python version 3"
-;;
+#!/bin/bash -xe
+
+case ${1} in
```
gucharbon: It may not be worth the effort, but I think that if you create a docker image for jarvis, it should not use the `setup.sh`  script since environment is well controlled compared to bare metal machine. 

Instead, iI suggest you install deps in Dockerfile, and provide several Dockerfiles if you wish to support several distributions (which may not be necessary). This way you will reduce drastically image size (no virtualenv and no cache when installing with apt or pip).

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN useradd -Um jarvis && echo "jarvis  ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+RUN chown -R jarvis /app
+USER jarvis
+
+RUN sudo /bin/bash ./setup.sh --python3
```
gucharbon: It may not be worth the effort, but I think that if you create a docker image for jarvis, it should not use the setup.sh script since environment is well controlled compared to bare metal machine.

Instead, iI suggest you install deps in Dockerfile, and provide several Dockerfiles if you wish to support several distributions (which may not be necessary). This way you will reduce drastically image size (no virtualenv and no cache when installing with apt or pip, remove archives when installing phantomjs, etc..).

Dockerfile
```diff
@@ -0,0 +1,58 @@
+FROM python:3.6.5-jessie
+
+env DEBIAN_FRONTEND noninteractive
```
gucharbon: Docker linters do not recognize env if it's in lower case most of the time. If you can just change to 
`ENV DEBIAN_FRONTEND="noninteractive"`
it would be more readable

Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+            xdg-utils \
+            xorg  && \
+     rm -rf /var/lib/apt/lists/*
+
+
+WORKDIR /usr/local/share
+RUN wget -qO - https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 | tar xjf - && \
+  ln -s /usr/local/share/phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/local/bin/phantomjs
+
+COPY . /app
```
gucharbon: I'm not sure if this will set recursively, but you can set ownership during copy:
`COPY --chown=jarvis:jarvis . /app`

A big advantage is that since docker layers are read only, when copying first and then setting permission, you add twice the size of the content you copy. We noticed that when building image with pretty large content (>1GB) and when modifying ownership in another layer it added 1 GB in the image size. See https://medium.com/@lmakarov/the-backlash-of-chmod-chown-mv-in-your-dockerfile-f12fe08c0b55 and https://github.com/moby/moby/pull/34263#issuecomment-343963040

Dockerfile
```diff
@@ -1,33 +1,58 @@
...
+RUN echo 'deb http://deb.debian.org/debian jessie-backports contrib non-free main' >> /etc/apt/sources.list && \
+    apt-get update && \
+    apt-get -y install --no-install-recommends \
+            build-essential \
+            chromedriver \
+            chrpath \
             dbus-x11 \
+            ffmpeg \
             libdbus-1-dev \
+            libdbus-glib-1-dev \
```
gucharbon: You start from python image with version 3.6.5 installed but this line install python3.5 as dependency. You could start from debian instead of python if you do not want to have several python installed (increase size for nothing). You can check that with 
`apt install --dry-run  libdbus-glib-1-dev`

Dockerfile
```diff
@@ -1,33 +1,58 @@
...
+            libffi-dev \
+            libfontconfig1 \
+            libfontconfig1-dev \
+            libfreetype6 \
+            libfreetype6-dev \
             libnotify-cil-dev \
+            libssl-dev \
+            libxft-dev \
             notification-daemon \
+            python3-dbus \
```
gucharbon: Like before, either build libraries which need python from source or use debian base image instead of python. This will install python3 as debian still uses python3.5 instead of python3.6 by default

Dockerfile
```diff
@@ -1,33 +1,58 @@
...
+            python-dbus \
```
gucharbon: Remove this line, it will install python2, I don't think you need it if you are using python3.
 The same goes for `python-notify2` . You can install `python3-notify2` instead. 
`python-imdbpy` will work for python3 you can keep this.

Dockerfile
jpigree: Oh. I did not realize this. Thanks. This should decrease the image size.

Dockerfile
jpigree: > > Hi @gucharbon, thank you for your review. I am working on the fixes you asked. I just have issues making the image work again (I rebased on top of the mainline). I will keep you updated.
> 
> I mentioned two little things that could still be changed, but you did a really nice job with this new Dockerfile !
> 
> But in any case I'm not reviewer or maintainer of this project, so even if I review it you will still need approval of a maintainer :)

Thanks even more for the review. But how can you be assigned as a reviewer of this PR if you are not a maintainer/reviewer of the project?

jpigree: Oh Okay, I missread the Github interface. It just says that you reviewed my PR but you are not assigned.

StefanosChaliasos:
@jpigree When you are ready please update the README and publish the image to [docker hub](https://hub.docker.com/). It would be great if you could maintain the image.

Dockerfile
gucharbon: Yep, just Killing some time :)

pnhofmann:
So - what's the progress?

Noticed some dependencies (chromedriver, python3-dbus, build-essential, phantomjs) are obsolete by now (refere #336).

I don't really understand why sudo is configured. Can't think of any Jarvis-functionality, which requires root.

But furthermore, this Dockerfile looks ready for me.

aditisrinivas97:
@pnhofmann sudo needs to be configured because the setup.sh file uses it on line 70. If its not configured, you get the following error message : 

`./setup.sh: line 70: sudo: command not found`

pnhofmann:
Setup.sh isn't used for Docker.

At last in the latest version of this Dockerfile - sudo looks like leftover from first version which did call setup. Especially since the added sudo-line is removed with 'sed' 2 lines below.

aditisrinivas97:
Right! It's in the first version. Thanks!

jpigree:
Hi. @StefanosChaliasos I just saw your comment today. Sorry. It has been a long time.
Are you still interested into putting this in dockerhub? If I do it myself, this means I will create (and own) the Dockerhub organization "Jarvis" then create the repository for the Docker image. I think this would be unwise as I am not even a contributor of the project yet. This is my first contribution and I don't even use Jarvis for now. What I can do is do all the setup work then hand it over a collaborator/admin of this repository.

@pnhofmann this PR was done in May 2018. I just looked at the history, a lot of commits have been merged since so it needs to be fully retested and adjusted. I will also look at the sudo matter. Give me a few days to rework this.

Hope it will be merged soon afterwards. Don't want to redo the same work again and again. :)

StefanosChaliasos:
@jpigree Hello, please create a Dockerfile for the last commit on master branch. We remove most of the dependencies so it's must be simpler.

jpigree:
I updated the Dockerfile. However, from what I see,

@StefanosChaliasos  Jarvis does not require the service X anymore?
If so, I will remove the X socket mount in the Makefile command "run_docker".
Then we should be good. We can merge it. For Dockerhub, somebody wants to be owner? I think it will be best to have multiple owners.

StefanosChaliasos:
@jpigree you can push it on Dockerhub

PR #325, opened 2018-06-19
Error checks in do_wiki() function for "wiki" command.

kaynaat007:
The command wiki was crashing after supplying no arguments to "wiki"
command. Also since wiki command always needs arguments after it's first arg, checked for index at 1.


appi147:
Wiki has moved away from cmdinterpreter, please reopen PR from updated codebase

PR #338, opened 2018-09-27
Update Jarvis.py

CSEian123:
optimized


appi147:
I don't know how but it breaks feature

appi147:
Reopen with updated master if solved

Issue #339, opened 2018-10-02
"remind" broken?


pnhofmann:
```
~> What can i do for you?
remind add 14:25 buy tomatoes
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/tmp/Jarvis/jarviscli/__main__.py", line 11, in <module>
    main()
  File "/tmp/Jarvis/jarviscli/__main__.py", line 7, in main
    jarvis.executor()
  File "jarviscli/Jarvis.py", line 173, in executor
    self.cmdloop(self.first_reaction_text)
  File "/usr/lib64/python2.7/cmd.py", line 142, in cmdloop
    stop = self.onecmd(line)
  File "/usr/lib64/python2.7/cmd.py", line 221, in onecmd
    return func(arg)
  File "jarviscli/CmdInterpreter.py", line 603, in do_remind
    reminder_handler(data)
  File "jarviscli/utilities/GeneralUtilities.py", line 62, in wrapped
    func(*args, **kwargs)
  File "jarviscli/packages/reminder.py", line 190, in reminder_handler
    globals()[action](data)
KeyError: 'handlerAdd'
```

Apparently this is just a simple typo - looks like during Reflactoring "handlerAdd" was renamed "handler_add" - without renaming in `` addAction("handlerAdd", ["add", "new", "create"], minArgs=1)``.

The interesting part: This file was last modified about a year ago (19. Oct 2017). So I guess, probably nobody is using this feature? 

Did implement something to send notifications without python-dbus for #336 - but I noticed other problems:

* remind (and also todo) should use jarviscli.packages.memory.memory instead writing their own files
* Don't really like the way the command is processed (``A match is determined by similarity between request and the available reminder names.`` Processing like that should be implemented on a application-level - and not in one command. There are interesting libraries (https://www.nltk.org/, https://rustyonrampage.github.io/text-mining/2017/11/28/spelling-correction-with-python-and-nltk.html). In fact honestly I already planted to take a closer look at those - but guess that's another story ;).
* Want to write something like "remind me in 10 minutes to take pizza out of oven"

=> Anyone would mind if I completely re-implement this module (and probably also parts of todo)? Any suggestions / thoughts?

StefanosChaliasos:
Go for it @pnhofmann !

Issue #336, opened 2018-09-13
Remove dbus and other non python dependencies


StefanosChaliasos:
We should consider to make jarvis more universal and more maintainable. I propose to remove dependencies like dbus and refactor the code base to work with only python packages. All the active members of this project are welcome to participate in this discussion. @appi147 @pnhofmann @sukeesh @kaynaat007 @jpigree @gucharbon

StefanosChaliasos:
In the first place we could remove those features and then find replacements to implement them again. The installation process should be universal on any os with just a pip install. I will create a new branch for this issue do you want to work on it @pnhofmann ?

appi147:
@StefanosChaliasos Instead of removing we could add wrappers like IS_DEBIAN. e.g. IS_MACOS implemented by @danielcorin  [here](https://github.com/sukeesh/Jarvis/blob/4bf7aa3e2e7e6bac3ae1001aa7305c0458b0804f/jarviscli/CmdInterpreter.py#L15).

In long term, IMO e should remove all package manager dependencies

pnhofmann:
Thanks for collaborator invitation ;) !
I'll see what I can do.

hugolesta:
Hello there,

After install Jarvis the installation returns the following error:

```
    checking for Python include path... -I/usr/include/python2.7
    checking for Python library path... -L/usr/lib -lpython2.7
    checking for Python site-packages path... /usr/lib/python2.7/dist-packages
    checking python extra libraries... -lpthread -ldl  -lutil -lm
    checking python extra linking flags... -Xlinker -export-dynamic -Wl,-O1 -Wl,-Bsymbolic-functions
    checking consistency of all components of python development environment... no
    configure: error: in `/tmp/pip-build-Go1dr4/dbus-python/build/temp.linux-x86_64-2.7':
    configure: error:
      Could not link test program to Python. Maybe the main Python library has been
      installed in some non-standard library path. If so, pass it to configure,
      via the LIBS environment variable.
      Example: ./configure LIBS="-L/usr/non-standard-path/python/lib"
      ============================================================================
       ERROR!
       You probably have to install the development version of the Python package
       for your distribution.  The exact name of this package varies among them.
      ============================================================================

    See `config.log' for more details
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-Go1dr4/dbus-python/setup.py", line 109, in <module>
        tests_require=['tap.py'],
[...]

PR #341, opened 2018-10-07
Python dependencies only

pnhofmann:


pnhofmann:
Patches for #336. Please take a look.
Also includes new remind-command #339 - since old remind command needs dbus.

Please take a look! Especially - since I can't do that - someone test "remind" on MacOS - and check if the notification works as expected.

StefanosChaliasos:
@pnhofmann Did you check all the tests? If you want help with tests you can open an issue.

pnhofmann:
@StefanosChaliasos 
Thanks!
Fixed music test and some code still issues.
The timeIn Test failure has nothing to do with these patches - in fact this test case even fails on with "beforepluginmanager". Created issue #342

appi147:
Is it ready @pnhofmann ?

pnhofmann:
@appi147 
Mostly. In fact, I have only some thought about setup.sh

The setup.sh on Mac writes 'python' or 'python3' to jarvis depending if python2 or python3 is used. Is this really necessary - since on Linux with python3 virtualenv python is symbolic link to python3. Does virtualenv really differ between Mac and Linux? Because if not, we could remove the whole os-specific code in setup.sh and just check in
```
#!/bin/bash
source /home/phil/projects/Jarvis/env/bin/activate
python /home/phil/projects/Jarvis/jarviscli/
```
as jarvis.
 
@StefanosChaliasos 
About
``` cp jarvis /usr/local/bin```
On most systems we would probably need ``sudo``. Don't really like doing anything outside /home in setup.sh.

I would propose asking the user:

```
echo "How to start Jarvis?"
echo "1: Call full path $jarvispath/jarvis"
echo "2: Copy jarvis to /usr/local/bin (root required)"
echo "3: Add $jarvispath to \$PATH"
read -n 1 -p "(Default: 1) " answer
echo ""
echo ""
echo "Installation complete. Try using Jarvis!"
case ${answer:0:1} in
2 )
    sudo cp jarvis /usr/local/bin
    echo "\$ jarvis"
;;
3)
    echo "export PATH=\$PATH:$jarvispath" >> ~/.bashrc
    echo "\$ jarvis"
;;
* )
    echo "\$ $jarvispath/jarvis"
;;
esac
```

**Comment by appi147** [comment]
Don't know about  MacOS, but it would be good. MacOS comes with python2 as default python IMO(depends on version of MacOS). Adding $jarvispath to path would be a good option. Don't know if we can set path same in MacOS.

**Comment by appi147** [comment]
OK here are tests:
- remind works(sent a small notification)
- remind remove crashes Jarvis if there is no reminder(empty reminder list)
- facebook and twitter are not present(as seen in commits)
- sound is fine
- hackathon is fine
Python2 is fine

**Comment by pnhofmann** [comment]
@appi147
Thanks! Fixed reminder crash.

**Comment by pnhofmann** [comment]
Ok if I rebase master or is a merge prefered?

**Comment by appi147** [comment]
Rebase pls

File: setup.sh [hunk 1]
```diff
@@ -1,4 +1,18 @@
 #!/bin/bash
+# fail on error!
+set -euo pipefail
```
appi147: [https://github.com/pypa/virtualenv/issues/1029](https://github.com/pypa/virtualenv/issues/1029)
`source /env/bin/activate` fails on my machine

setup.sh
```diff
@@ -40,98 +54,20 @@ exec 3<> jarvis
...
   ;;
 esac
 
 source env/bin/activate
-pip install -r requirements.txt
-pip install dbus-python
+
+pip install --upgrade -r requirements.txt
+
+cp jarvis /usr/local/bin
```
appi147:
```suggestion
sudo cp jarvis /usr/local/bin
```

appi147:
@pnhofmann

appi147:
Excellent work @pnhofmann

PR #343, opened 2018-10-09
 Shutdown: Add suspend, hibernate, hybridsleep

pnhofmann:
Reboot, Shutdown: Only ask "In how many minutes" if rest of command
is empty

### DIFF
--- a/jarviscli/plugins/shutdown.py
+++ b/jarviscli/plugins/shutdown.py
@@ -5,9 +5,10 @@ from plugin import plugin
 
 @plugin()
 def shutdown(jarvis, s):
-    """Shutdown the system."""
-    minutes = input('In how many minutes?: ')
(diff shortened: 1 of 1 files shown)


appi147:
@pnhofmann Please add `shutdown -c` option to it. e.g.
```
~> Hi, what can I do for you?
shutdown
In how many minutes?: 1
[sudo] password for appi: 
Shutdown scheduled for Thu 2018-10-11 16:33:11 IST, use 'shutdown -c' to cancel.
~> What can i do for you?
shutdown -c
Shutdown scheduled for Thu 2018-10-11 16:33:16 IST, use 'shutdown -c' to cancel.
```

appi147:
I see that shutdown cancel is implemented, but as in above example typing `shutdown -c` schedules shutdown after 1 min

PR #344, opened 2018-10-09
More maths with sympy

pnhofmann:
* Calculate with sympy instead use 'eval'
* variable support (x, y)
* Plot graphs
* Solve equations
* Solve Linear systems
* Factories term
* Limit
* Curve Sketching

### DIFF
--- a/jarviscli/plugins/evaluator.py
+++ b/jarviscli/plugins/evaluator.py
@@ -1,11 +1,14 @@
-from math import *  # to give eval access to all of math lib
+import re
+import sympy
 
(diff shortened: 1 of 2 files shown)


pnhofmann:
Got a book about Math with Python - and just had to "play" a bit ;). Like that SymPy library!

Please tell, if this is too much - and I'll remove overdone features.

appi147:
Nice work!! I was unable to use `equations`, how to use it?

pnhofmann:
Thanks! Added equations example.

PR #347, opened 2018-10-13
Bugfix

pnhofmann:


pnhofmann:
Really stupid bug with these relative / absolute paths and not loading plugins - should have noticed that earlier...
Additionally made auto-complete work with two word commands - somehow the iterator.next() implementation did not work and my implementation for plugins was also sort of broken.

Issue #342, opened 2018-10-09
"Check time in" not working


pnhofmann:
```
~> Hi, what can I do for you?
check time in Greenwich, UK
Traceback (most recent call last):
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/phil/projects/Jarvis/jarviscli/__main__.py", line 11, in <module>
    main()
  File "/home/phil/projects/Jarvis/jarviscli/__main__.py", line 7, in main
    jarvis.executor()
  File "/home/phil/projects/Jarvis/jarviscli/Jarvis.py", line 173, in executor
    self.cmdloop(self.first_reaction_text)
  File "/usr/lib64/python3.6/cmd.py", line 138, in cmdloop
    stop = self.onecmd(line)
  File "/usr/lib64/python3.6/cmd.py", line 217, in onecmd
    return func(arg)
  File "/home/phil/projects/Jarvis/jarviscli/CmdInterpreter.py", line 299, in do_check
    timeIn.main(self, s)
  File "/home/phil/projects/Jarvis/jarviscli/packages/timeIn.py", line 12, in main
    loc = getLocation(s)
  File "/home/phil/projects/Jarvis/jarviscli/packages/timeIn.py", line 34, in getLocation
    lat = j['results'][0]['geometry']['location']['lat']    # Latitude
IndexError: list index out of range
```
Guess the API changed...

appi147:
google maps api is giving this error
```
{
   "error_message" : "You have exceeded your rate-limit for this API. We recommend registering for a key at the Google Developers Console: https://console.developers.google.com/apis/credentials?project=_",
   "results" : [],
   "status" : "OVER_QUERY_LIMIT"
}

```
thus list index out of range error

appi147:
Working correctly with key
```
17:47 $ python timeIn.py 
{
   "results" : [
      {
         "address_components" : [
            {
               "long_name" : "Greenwich",
               "short_name" : "Greenwich",
               "types" : [ "neighborhood", "political" ]
            },
            {
               "long_name" : "London",
               "short_name" : "London",
               "types" : [ "postal_town" ]
            },
            {
               "long_name" : "Greater London",
               "short_name" : "Greater London",
               "types" : [ "administrative_area_level_2", "political" ]
            },
            {
[...]

PR #355, opened 2018-10-18
fix time in

appi147:
#342


appi147:
Can you review please @pnhofmann ?

jarviscli/packages/timeIn.py
```diff
@@ -27,11 +39,22 @@ def main(self, s):
...
-    lat = j['results'][0]['geometry']['location']['lat']    # Latitude
-    lng = j['results'][0]['geometry']['location']['lng']    # Longitude
-    # Returns both latitude and longitude as a tupple
-    return lat, lng
+    try:
+        lat = j['results'][0]['geometry']['location']['lat']    # Latitude
+        lng = j['results'][0]['geometry']['location']['lng']    # Longitude
+        # Returns both latitude and longitude as a tupple
+        return lat, lng
+    except IndexError:
```
pnhofmann:
```suggestion
    except IndexError r:
```
Actually very reasonable asking to open an github issue instead hard of crashing ;). But suppose it would be better to to try-except every Error in CmdInterpreter before calling any command.

jarviscli/packages/timeIn.py
```diff
@@ -1,16 +1,28 @@
...
+# this sets the path to the modules directory not the directory it was call from
+module_path = os.path.dirname(__file__)
+
 
 def main(self, s):
     # Trims input s to be just the city/region name
     s = s.replace('time ', '').replace('in ', '')
+    exists = os.path.isfile(module_path + '/../data/keys.json')
+    if not exists:
+        shutil.copy2(module_path + '/../data/samplekeys.json', module_path + '/../data/keys.json')
```
pnhofmann: Don't think we should do that here - more API keys might come and that would duplicate this code. Possible places would probably be:
* setup.sh
* ``__main__.py``

jarviscli/packages/timeIn.py
```diff
@@ -27,11 +39,22 @@ def main(self, s):
 
 
 def getLocation(s):
-    send_url = ("http://maps.googleapis.com/maps/api/geocode/json?address={0}&sensor=false".format(s)
-                )
+    file_path = module_path + '/../data/keys.json'
+    with open(file_path) as json_file:
+        data = json.load(json_file)
+    key = data['timein']
```
pnhofmann:
```suggestion
    if 'timein' not in data or data['timein'] == 'insertyourkeyhere':
        print(Fore.RED + "Generate api key here: https://developers.google.com/maps/documentation/geocoding/start?hl=en_US")
        print(Fore.RED + "and add it to jarviscli/data/keys.json" + Fore.RESET)
        return None
    key = data['timein']
```

Would check if key is "insertyourkeyhere".

Probably not the "cleanest" solution but imagine user runs "time in" once, forgets / is too lacy to creating API key but runs "check time in" again - he would be asked to open issue instead creating an API key.

Would also replace os.path.isfile check from above.

jarviscli/packages/timeIn.py
pnhofmann: Forget this one ;). I'll create a pull request some day, because it's a different topic.

appi147:
my idea of samplekey file was to create a json for every api key a user need to register for

jarviscli/packages/timeIn.py
appi147: agree with this one, I thought of it but was unable to implement it in a nice way :(

pnhofmann:
Oh yes, you are right. Haven't thought about updating the keys.json file for new entries - that's probably ugly and not worth the effort. Agree, one key file for every api key is reasonable.

Could you just rename keys.json and samplekeys.json to something like keys_googlemapps.json and samplekeys_googlemapps.json?

PR #357, opened 2018-10-20
Jarvis Windows Port!!

KrDimitrov:
ported the setup bash script to windows

Jarvis installs as usual

**The jarvis directory is added to The users path variable (not `%PATH%`)**, which means that jarvis **cannot be run from other users** using a terminal, either run `setup.bat` again or just run ` jarvis.bat` in the jarvis directory

**Jarvis's output is messy since i have not done any work on any python files**

No missing packages are reported


pnhofmann:
Merged into branch [windows port](https://github.com/sukeesh/Jarvis/tree/windows-port).

Aiki-bot51:
why when i clone there is no setup.bat

appi147:
@Aiki-bot51 Check windows branch

Issue #301, opened 2018-03-20
Create Dockerfile


StefanosChaliasos:

StefanosChaliasos:
@SolaOmi go for it. It is straightforward what you should do. Just create a Dockerfile so one can create a container to use jarvis.

StefanosChaliasos:
@SolaOmi any updates?

akashshah26595:
@StefanosChaliasos Can I help to fix this issue? I've created few Docker files before.  It won't take too long.

StefanosChaliasos:
I think @SolaOmi works on it.

SolaOmi:
@StefanosChaliasos @akashshah26595 I'm having some trouble with the [dockerfile](https://gist.github.com/SolaOmi/831160dacf59cf4471b3d9f8cc56b9f1), it breaks down during the build when I get to the RUN ./setup.sh step, and returns this error:

"The command '/bin/sh -c ./setup.sh' returned a non-zero code: 1"

thoughts?

StefanosChaliasos:
@akashshah26595 Do you want to help?

akashshah26595:
@SolaOmi @StefanosChaliasos Try [this](https://gist.github.com/akashshah26595/1a7d4dd52e2fb51f02e5c3c8e74196bb) Dockerfile.

akashshah26595:
Tested this Dockerfile. Container gets created launching bash, having files in /app directory, executing setup.sh and exposing port 80.

StefanosChaliasos:
@SolaOmi any updates? @akashshah26595 do you want to create a PR with your DockerFile?

akashshah26595:
- @StefanosChaliasos Can you provide more details on what this Dockerfile should do?
- It would help me to design Dockerfile accordingly.

StefanosChaliasos:
Just create a Dockerfile that one could use it to run jarvis (interactive).

jpigree:
Hello guys. I am motivated to take the task (I want to give it a try but inside a Docker anyway). I have a Dockerfile in which I installed Jarvis. I am debugging execution time issues at the moment.

jpigree:
@SolaOmi, @akashshah26595 can I take the issue?

StefanosChaliasos:
@jpigree please create a PR when the Dockerfile is ready.

jpigree:
Okay. Thank you @StefanosChaliasos. I have still have dbus issues at the moment. I am working on it.

[...]

PR #364, opened 2018-10-29
Add Dockerfile - Issue  #301

gabrielperes97:
I've created a simple dockerfile. It runs, but the external commands such volume and hotspot control not works yet, need some OS binds for it.


pnhofmann:
Closing in favor of #318

PR #371, opened 2018-11-04
Error message when No cricket matches are being played

tmsreekanth98:
Earlier, when the pycricbuzz library was returning an empty list of cricket games, jarvis was stuck in an infinite loop requesting for a choice. Now an error message of "No Matches Being Played!" has been added and the control returns immediately from the respective function when such a situation occurs. This change improves user experience.


pnhofmann:
So since you opened #372, I suppose this PR is obsolete? Closing therefore.

You actually don't have to open a new PR if you add something - just push new commits. You can even overwrite old commits using [git --force](https://stackoverflow.com/questions/10510462/force-git-push-to-overwrite-remote-files).

tmsreekanth98:
Sorry, I am new to open source and git. I had a question about the PR. Why
was it showing Travis build failed? I have added only 3 lines of code.
Thanks

pnhofmann:
No problem - I'm quite new, too ;) As noted under your new PR - just ignore Travis for now ;).

PR #372, opened 2018-11-04
Added error message for no cricket matches being played

tmsreekanth98:
Creating new pull request because the earlier one shows travis ci build failed. If it still fails the build check please let me know the reason for the failure because i am quite new to all this.

Earlier, when the pycricbuzz library was returning an empty list of cricket games, jarvis was stuck in an infinite loop requesting for a choice. Now an error message of "No Matches Being Played!" has been added and the control returns immediately from the respective function when such a situation occurs. This change improves user experience.

### DIFF
--- a/jarviscli/plugins/cricket.py
+++ b/jarviscli/plugins/cricket.py
@@ -115,6 +115,9 @@ class Cricket(Plugin):
     def score(self, jarvis):
         matches = self.all_matches()
         jarvis.say(Fore.RED + "\nALL MATCHES\n" + Fore.LIGHTBLUE_EX)
+        if matches == []:
+            jarvis.say(Fore.BLACK + "No Matches Being Played!\n" + Fore.RESET)
+            return
         for i, m in enumerate(matches, 1):
             jarvis.say("{}. {}".format(str(i), m))
         choice = int(input(Fore.RED + '\nEnter choice (number): ' + Fore.RESET))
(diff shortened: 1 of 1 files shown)


pnhofmann:
> Creating new pull request because the earlier one shows travis ci build failed. If it still fails the build check please let me know the reason for the failure because i am quite new to all this.

Just ignore ;).

This is no real "build" failing but rather a notification, that code "flaws" were found. Means, test failure or something flake (auto checks code style) reported.
You can check this by calling
``source env/bin/activate``
``./test.sh``

So why the errors?

A new version of flake8 was released reporting some new code style issues, it didn't report earlier. Nothing to do with your changes ;).

tmsreekanth98:
Thanks for the reply. I thought the modifications were having some problem.

jarviscli/plugins/cricket.py
```diff
@@ -115,6 +115,9 @@ def scorecard(self, desc):
     def score(self, jarvis):
         matches = self.all_matches()
         jarvis.say(Fore.RED + "\nALL MATCHES\n" + Fore.LIGHTBLUE_EX)
+        if matches == []:
+            jarvis.say(Fore.BLACK + "No Matches Being Played!\n" + Fore.RESET)
```
pnhofmann: Black is a quite sub-optimal color - since black or really dark colors are quite common for terminal backgrounds.

**Thanks for fixing this bug!**



By the way; jarvis.say() also has an option for color:
```suggestion
            jarvis.say("No Matches Being Played!\n", Fore.RED)
```
But this option isn't used in this whole module anyway ;).

tmsreekanth98:
No problem. :) Happy to fix the bug. Black looked fine while testing that's
why I used it.

PR #375, opened 2018-11-05
bug wiki fixed

Jamparan:
I already solve the bug on wiki command. If someone write the command wrong jarvis will give a few possible options to work with.


pnhofmann:
Thanks!

PR #396, opened 2019-01-19
Enhanced output of os plugin on macOS

bsushmith:
Before:
screenshot 2019-01-19 at 8 12 10 pm

After:
screenshot 2019-01-19 at 8 07 04 pm

### DIFF
--- a/jarviscli/plugins/systemOptions.py
+++ b/jarviscli/plugins/systemOptions.py
@@ -1,9 +1,7 @@
 import os
-from platform import architecture, dist, release
+from platform import architecture, dist, release, mac_ver
 from platform import system as sys
-
-from colorama import Fore
-
+from colorama import Fore, Style
 from plugin import LINUX, MACOS, PYTHON2, PYTHON3, plugin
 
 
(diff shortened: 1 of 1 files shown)


bsushmith:
There are old commits (some of which were already merged with this repo in previous pull requests)that have come into this pull request. Still new to this... can someone help me point in the right direction on how to avoid this. Thanks in advance

pnhofmann:
Thanks!

> There are old commits (some of which were already merged with this repo in previous pull requests)that have come into this pull request. Still new to this... can someone help me point in the right direction on how to avoid this. Thanks in advance

You are using git cli client? So basically it's about **rebase**.

Best read about everything git offers in the Book "Pro Git" which you can read online for _free_ [here](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) (Creative Commands 3.0).

Basically rebase re-commits all commits on top of another branch.

If you take a closer look, for example your commit "update curl command to use --connect-timeout" here is called "cc549a8". The commit merged into master is called "85d547b". I merged your PR with the option "Rebase and merge" - that created **new** commit which looks like your original commit, has the same content, but is a completely different commit for git.

So that's the reason why old commits how up here. So what to do?

* Never merge your old branches; if they are merged, you can just delete them
* Always go "from master"
* Never do changes on master or merge something to master
* On master, only do ``git pull origin``

Before submitting a PR, you should do something like that:
```
git checkout master
git pull
git checkout YOUR_BRANCH
git rebase master 
```

Easy as that! You could also do a "git merge master", but git rebase is just more elegant, because it neither creates a merge commit nor does it show up in git log.

So first, guess you need a clean origin/master.

```git checkout -B master origin/master```

The do rebase as noted above (checkout master, pull, checkout branch, rebase master). And - easy as that!

Note that you need to do ``git push -f`` after you did a rebase. That's because rebase re-creates new commits and you need to overwrite the old versions (even if no line changed).

Also note to never rebase after someone pulled your commits and commitet atop - that creates a mess ;).

So - just try it out and check if it works!

bsushmith:
Thank you @pnhofmann. The documentation in the Pro Git book, helped me to see git from a different perspective than i saw before. I'll try this to come out in the proper way. Menahwile, I'll close this pull request and get things cleaned out.

PR #397, opened 2019-01-20
enchanced output of os plugin on macOS

bsushmith:
Before:
screenshot 2019-01-19 at 8 12 10 pm

After:
screenshot 2019-01-19 at 8 07 04 pm


pnhofmann:
> Thanks for your help earlier.

No problem. Looks good! I'll just merge. Thanks again for your contribution!

PR #399, opened 2019-01-21
Add network dependency to plugins which require it

bsushmith:
The changes in this PR are as follows - 

-  Add network dependency to the plugins
     - For lyircs & currencyconv plugins, add network in require function

     - For check forecast & speedtest, add a try block to return connection_error() msg in case of no network

     - for gmail, return instead of exit if no network

### DIFF
--- a/jarviscli/plugins/speedtest.py
+++ b/jarviscli/plugins/speedtest.py
@@ -38,7 +38,7 @@ def speedtest(jarvis, s):
     """Runs a speedtest on your internet connection"""
     try:
         res = st.Speedtest()
-    except:
+    except st.ConfigRetrievalError:
         return jarvis.connection_error()
 
     # Create a spinner on command line to show that its running
--- a/jarviscli/plugins/speedtest.py
+++ b/jarviscli/plugins/speedtest.py
@@ -36,8 +36,10 @@ class SpinnerThread(threading.Thread):
 @plugin(network=True)
 def speedtest(jarvis, s):
     """Runs a speedtest on your internet connection"""
-
-    res = st.Speedtest()
+    try:
+        res = st.Speedtest()
+    except:
+        return jarvis.connection_error()
 
     # Create a spinner on command line to show that its running
     spinner = SpinnerThread('Running the test ', 0.15)
--- a/jarviscli/CmdInterpreter.py
+++ b/jarviscli/CmdInterpreter.py
@@ -272,7 +272,10 @@ class CmdInterpreter(Cmd):
         elif "time" in s:
             timeIn.main(self, s)
         elif "forecast" in s:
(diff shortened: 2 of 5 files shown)


jarviscli/plugins/speedtest.py
```diff
@@ -36,8 +36,10 @@ def stop(self):
 @plugin(network=True)
 def speedtest(jarvis, s):
     """Runs a speedtest on your internet connection"""
-
-    res = st.Speedtest()
+    try:
+        res = st.Speedtest()
+    except:
```
pnhofmann: As described by this code review tool:
https://app.codacy.com/app/StefanosChaliasos/Jarvis/pullRequest?prid=2920527
except-ing all kind of exceptions is usually not best idea, because you never know: If "You are not connected to Internet" shows up; is it really because of a connection error or because of another error we didn't think off?

Of course this is a really **small** issue, but nevertheless would recommend changing this line to (you should be able to apply with one click):

```suggestion
    except st.ConfigRetrievalError:
```

In addition, thanks again for you contribution!

bsushmith: I was having some network problems the day I was testing it, and the specific exception didn't seem to work. It seems to be working fine now, so I will apply your suggestions.

pnhofmann:
Thanks!

bsushmith:
Thank you!
I have one quick question - 
Before this merge, my fork was exactly two commits ahead. 
After the merge, my fork is "This branch is 2 commits ahead, 2 commits behind sukeesh:master"

When the same thing happened before, what I did was - 
```
git checkout master
git fetch upstream
git reset --hard upstream/master
git push --force
```

Is this a common recurrence?

pnhofmann:
> Is this a common recurrence?

Does definitely work.
Would recommend using

```git checkout -B master origin/master```

instead

```git reset --hard upstream/master```

because reset --hard is so destructive. Most git commands only add data so you can recover anything and undo any command. ``git reset --hard`` is one of the exceptions which destroys data; so that might result in data loss if you make a mistake (e.g. any uncommitted changes will be lost after reset --hard). So better avoid ;).

So anyway:

Better don't merge branches you submitted via Pull Request to your own master if you have no good reason - there is no point and it just messes up git ;). Better pull upstream after your branch was merged.

In fact, you don't actually have to maintain your fork. I never push to master of forked projects. A fork, which is just there to push commits before creating a Pull Request may be as well be lots of commits behind. But of course keeping your github-master up to date is absolutely fine; everyone has his/her own "workflow" ;).

pnhofmann:
Should also note, that I'm relatively new to github, too. So please don't take something I say as something like "absolute wisdom" ;).

bsushmith:
Yeah 👍

PR #400, opened 2019-01-23
Fix movie plugin for dict errors & update movie output

bsushmith:
- Refactor movie plugin code for more meaningful variable names and output
- search for presence of attribute in movie data dictionary before trying to fetching the value
- update movie plot output
- update movie info output

### DIFF
--- a/jarviscli/plugins/movie.py
+++ b/jarviscli/plugins/movie.py
@@ -8,10 +8,10 @@ app = imdb.IMDb()
 
 
 def main(jarvis, movie):
-    movieId = search_movie(jarvis, movie)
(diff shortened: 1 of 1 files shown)


bsushmith:
Travis-CI check failed due to W504 error. I have updated the line break before the binary operator. 
But since the flake8 command line thats running is


`flake8 --select E,W --max-line-length=140 --ignore E722 jarviscli/`

Its catching the W503 error., which shouldnt be a problem.

pnhofmann:
Nothing to add. This is a real improvement of one of my favorite modules ;). Thanks!

PR #401, opened 2019-01-26
Add movie search plugin

bsushmith:
- Add movie search plugin which gives the top 5 results of a movie search and a prompt to get further info of search result
- Refactor existing movie.py code for better code re-usability

**movie search**

screenshot 2019-01-26 at 1 39 35 pm


pnhofmann:
Thanks!

PR #404, opened 2019-01-31
Added a health plugin

ApostPap:
The health plugin can tell you about your Body Mass Index(BMI) and daily caloric intake while also suggesting calories for losing weight and for putting on weight.
Call ''health bmi" for BMI or "health calories" for calories.

### DIFF
--- a/jarviscli/plugins/health.py
+++ b/jarviscli/plugins/health.py
@@ -8,6 +8,7 @@ def health_bmi(jarvis, s):
     It is a measure of body mass based on height and weight.
     Add metric height(cm) and weight(kg). No decimal weight for now.
     #Example: health bmi 182 86
+    ^Source: https://en.wikipedia.org/wiki/Body_mass_index
     """
 
     strings = s.split()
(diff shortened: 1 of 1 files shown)


pnhofmann:
Hi and thanks for your contribution!

Your code looks fine to me and this is definitely a useful feature. Just some "formalities":
Take a look below my post, there it stats: "All checks have failed".

Let's start with the "Travis CI build". Go to your Jarvis directory and run ``./test.sh``. Besides running some unit tests (which all pass) this will call flake8. This is a "code style checker". Basically it make stuff more easy for everyone if every code is formatted the same way.

For python there is [PEP 8](https://www.python.org/dev/peps/pep-0008/) which defines how Python code should look like. And flake8 checks if the code really looks like specified in PEP 8. All our code is formatted according to PEP 8 and this should also apply in the future...

Next, click on this link "details" next to "Codacy/PR Quality". Codacy does some static review on your code quality. Not every issue reported there is necessary a problem. But sometimes it'll notice quite useful stuff - e.g. in your case it noted that the variable "error" is unused and therefore could (or should) be removed.

So, please fix all errors returned by ``./test.sh`` (mostly misplaced line breaks or spaces), remove the "error=0" line and this PR can be merged!

pnhofmann:
Nice work, Thanks!

I'm Sorry, just noticed one small question: How to calculate BMI is well defined. But where is the calories calculation taken from? We should probably add a source somewhere. Also: This calculation isn't somehow copyright protected?

ApostPap:
Formulas are not usually copyright especially if it is not a part of a business process. That equation is used many websites.

Thank you for your help.

pnhofmann:
Thanks!

PR #407, opened 2019-02-02
Modified News Plugin to make sites as configurable.

jay4563:


jarviscli/plugins/news.py
```diff
@@ -58,22 +60,71 @@ def run(self, jarvis, s):
                 jarvis.say("I couldn't find news", Fore.RED)
         else:
             try:
+                if len(jarvis.get_data('news-settings')) > 0:
```
pnhofmann: Crashes if news are not configured:
```
  File "/home/phil/git/Jarvis/jarviscli/CmdInterpreter.py", line 143, in try_do
    do(s)
  File "/home/phil/git/Jarvis/jarviscli/plugin.py", line 90, in _plugin_run_with_network_error
    run_func(jarvis, s)
  File "/home/phil/git/Jarvis/jarviscli/plugins/news.py", line 63, in run
    if len(jarvis.get_data('news-settings')) > 0:
TypeError: object of type 'NoneType' has no len()
```

Probably missing a ``is not None`` check.

jarviscli/plugins/news.py
```diff
@@ -58,22 +60,71 @@ def run(self, jarvis, s):
...
+    def configure_news_site(self, jarvis):
+        news_site_list = jarvis.get_data(News.get_key())
+        # Check if no sites are configured
+        if news_site_list is None:
+            news_site_list = []
+        print("List of configured news sites are as below")
+        for idx, site in enumerate(news_site_list):
+            print("{} : {}".format(idx + 1, site))
+
+        print("Search www.newsapi.org for the list of news channels. Enter site-key")
```
pnhofmann: That's probably just newsapi.org - without the 'www'. Better to link directly to https://newsapi.org/sources.

jarviscli/plugins/news.py
```diff
@@ -58,22 +60,71 @@ def run(self, jarvis, s):
...
+        if del_from_list.lower() == "y":
+            print("Enter the site index to be deleted. If multiple sites, then enter index with comma seperated values")
+            index_to_del = str(input())
+            items_idx_to_del = index_to_del.split(',')
+            list_to_del = [news_site_list[int(item) - 1] for item in items_idx_to_del]
+            news_site_list = [item for item in news_site_list if item not in list_to_del]
+            if len(news_site_list) == 0:
+                self.is_news_site_configured = False
+        jarvis.update_data(News.get_key(), news_site_list)
+        jarvis.save()
```
pnhofmann: jarvis.save() does not exist... Memory save automatically after each add /  update / del - so no need to call save manually.

jarviscli/plugins/news.py
```diff
@@ -58,22 +60,71 @@ def run(self, jarvis, s):
...
+            print("{} : {}".format(idx + 1, site))
+
+        print("Search www.newsapi.org for the list of news channels. Enter site-key")
+        site_key = str(input())
+        if site_key is '':
+            print("Wrong input.")
+        elif site_key in news_site_list:
+            print("Site already added. Hence not updating the list.")
+        else:
+            news_site_list.append(site_key)
```
pnhofmann: Possible improvement (optional and not required):
Would it be possible to validate 'site_key' and print something like "news source XXX does not exit" if user entered something wrong?

jarviscli/plugins/news.py
```diff
@@ -1,5 +1,6 @@
 # !!! This uses the https://newsapi.org/ api. TO comply with the TOU
 # !!! we must link back to this site whenever we display results.
+# This has been edited for understanding simple git operations.
```
pnhofmann: Remove this line ;).

jay4563: Removed. :)

jarviscli/plugins/news.py
jay4563: Sounds correct. Implemented the check.

jarviscli/plugins/news.py
jay4563: Removed.

jarviscli/plugins/news.py
jay4563: Modified the link to correct one.

jarviscli/plugins/news.py
jay4563: None check has been implemented

pnhofmann:
Noticed, with python3 urllib throuws (unlike with python2) an exception on http 404 - so added small try-except. This difference between py2 and py3 is a bit ugly in this module ;).

So, finally merged.

Thanks for creating this feature!

PR #408, opened 2019-02-03
Added voice recognition feature.

mmkumr:
Added voice recognition feature. I have to update the documentation. I will update it tomorrow.


pnhofmann:
Voice recognition is obviously something really cool!

Unfortunately noticed, google-cloud-speech requires grpcio - which requires native C/C++ extension. This is something we want to avoid - with just python this application runs everywhere. Installing grpcio needs C-Compiler and python development headers installed.

Also installation of grpcio fails on my FreeBSD system - so couldn't try out yet; didn't investigate yet and will use my Linux-system later.

Noticed,

```
Speech recognition engine/API support:

    CMU Sphinx (works offline)
    Google Speech Recognition
    Google Cloud Speech API
    Wit.ai
    Microsoft Bing Voice Recognition
    Houndify API
    IBM Speech to Text
    Snowboy Hotword Detection (works offline)
```

What would you think about choosing another backend (as default) and maybe implement an option to use Google API optionally if installed?

jarviscli/plugins/stt.py
```diff
@@ -0,0 +1,71 @@
...
+                if pinger[0:5] == "go to":
+                    os.system("wmctrl -a " + pinger[7:])
+                elif pinger[0:9] == "workspace":
+                    pinger = pinger[11:len(pinger)]
+                    if pinger == 'one':
+                        pinger = 1
+                    os.system("wmctrl -s " + str(int(voice) - 1))
+                else:
+                    line = pinger
+                    jarvis = Jarvis.Jarvis()
```
pnhofmann: Just noticed, no need to create new Jarvis instance here - ``jarvis._jarvis`` should also work. Or we just add a ``eval(str)``-method to JarvisAPI.

mmkumr: Done.

mmkumr:
> Voice recognition is obviously something really cool!
> 
> Unfortunately noticed, google-cloud-speech requires grpcio - which requires native C/C++ extension. This is something we want to avoid - with just python this application runs everywhere. Installing grpcio needs C-Compiler and python development headers installed.
> 
> Also installation of grpcio fails on my FreeBSD system - so couldn't try out yet; didn't investigate yet and will use my Linux-system later.
> 
> Noticed,
> 
> ```
> Speech recognition engine/API support:
> 
>     CMU Sphinx (works offline)
>     Google Speech Recognition
>     Google Cloud Speech API
>     Wit.ai
>     Microsoft Bing Voice Recognition
>     Houndify API
>     IBM Speech to Text
>     Snowboy Hotword Detection (works offline)
> ```
> What would you think about choosing another backend (as default) and maybe implement an option to use Google API optionally if installed?

As per my knowledge `Google Speech Recognition` will be better. I have tried to run this on my Linux machine it works. fine.

hazzus:
> I have tried to run this on my Linux machine it works. fine.

I agree with @pnhofmann that cross-platform is main feature, non-python requirements make this project not so beatiful. I would suggest non-online backend also, still need some work during offline. Else looks sweet, as I tested.

pnhofmann:
So since this is a quite important feature, I'll wait a bit before merging.

@mmkumr
But you would be OK, if a) backend would be selectable and b) non-Google backend will automatically be choosen if google-required libs fail to install?

Anyone who would want to impement by chance ;) ? If not, would implement myself - probably weekend and you could review afterwards. Fine?

mmkumr:
Is there any other way by with I can execute shell commands and Codacy will not show the warning?

mmkumr:
> So since this is a quite important feature, I'll wait a bit before merging.
> 
> @mmkumr
> But you would be OK, if a) backend would be selectable and b) non-Google backend will automatically be choosen if google-required libs fail to install?
> 
> Anyone who would want to impement by chance ;) ? If not, would implement myself - probably weekend and you could review afterwards. Fine?
Ok, I will use `Microsoft Bing Voice Recognition` is it ok?

pnhofmann:
> Is there any other way by with I can execute shell commands and Codacy will not show the warning?

Don't think so. Just ignore ;). Flake8 should not fail - Codacy sometimes gives usefull tips + makes live easier to review PRs. They are right, if they say executing shell should usually be avoided. But for Jarvis - of course shell code in such places is perfectely ok. So just don't take this "check failed" to serious.  

> Ok, I will use Microsoft Bing Voice Recognition is it ok?

About this python devel: Finally could some testing - works really nice! Unfortunately:

```

  File "/home/phil/git/Jarvis/env/lib/python3.6/site-packages/speech_recognition/__init__.py", line 108, in get_pyaudio
    import pyaudio
ModuleNotFoundError: No module named 'pyaudio'

During handling of the above exception, another exception occurred:

  File "/home/phil/git/Jarvis/jarviscli/CmdInterpreter.py", line 143, in try_do
    do(s)
  File "/home/phil/git/Jarvis/jarviscli/plugin.py", line 116, in __run_method
    run_method(jarvis, s)
  File "/home/phil/git/Jarvis/jarviscli/plugins/stt.py", line 14, in hear
    with sr.Microphone() as source:
  File "/home/phil/git/Jarvis/env/lib/python3.6/site-packages/speech_recognition/__init__.py", line 79, in __init__
    self.pyaudio_module = self.get_pyaudio()
  File "/home/phil/git/Jarvis/env/lib/python3.6/site-packages/speech_recognition/__init__.py", line 110, in get_pyaudio
    raise AttributeError("Could not find PyAudio; check installation")
AttributeError: Could not find PyAudio; check installation
```

That's interessting - why wasn't PyAudio automatically installed on my system if it is required by speech_recognition?
Could you check if you somehow manually installed PyAudio and if we need to add this to requirements.txt?

Big problem is, PyAudio also requires native extension - but if it does not work without native extension then so it is. At last we would to make sure, that Jarvis could run also without voice recognition.
 
So, about Bing Voice Recognition - not only because of native extension (which we might anyway) but overall I like the idea of the user being able to choose the voice recognition provider. I'm happy for any vendor you add, but of couse you don't have to ;). And of course I'm ok about Bing Voice Recognition (also as first fallback) just like Google as default.

jarviscli/plugins/stt.py
```diff
@@ -0,0 +1,73 @@
...
+                    if pinger == 'one':
+                        pinger = 1
+                    num = str(int(pinger) - 1)
+                    subprocess.call(["wmctrl", "-s", num], shell=False)
+                elif pinger[0:4] == "open":
+                    pinger = pinger[5:]
+                    subprocess.call(["nohup", pinger], shell=False)
+                else:
+                    line = pinger
+                    jarvis = jarvis._jarvis
```
pnhofmann: Just testet again - Really work amazing!

Told you to use jarvis._jarvis here. This created a variable-conflict - if this line is executed a second time jarvis allready is an instance of Jarvis - which does not have the attribute "_jarvis". So should be renamed something like

```_jarvis = jarvis._jarvis
line = _jarvis.precmd(line)
stop = _jarvis.onecmd(line)
stop = _jarvis.postcmd(stop, line)
```

mmkumr: Ok. Fine. :) I will try to fix pyaudio problem.

mmkumr:
When I am trying to run it in my local machine it works but when Travis is running tests it is not installing the packages so it shows the error. I have modified the setup.sh checkout the file.

pnhofmann:
AFAIK Travis isn't using setup.sh. Anyway,

```sudo apt install libasound-dev portaudio19-dev python-pyaudio```

Not everyone uses Ubuntu. Once setup.sh checked the os and installed fitted packages for each platform. Spend some time to rewrite this - see #336.

Did some searching ... and actually found a python-only library for recording audio which works on linux, mac and windows!
https://github.com/bastibe/SoundCard

Could try to implement - somewhen...

=> For now would say, don't worry too much about this portaudio / PyAudio stuff. Your code works great. If you want to implement anything else (portaudio alternative, different recognition backend, ...) you can (in this case please tell me) but you don't have too.

Oh yes, just one final thing I noticed:

```
                if pinger[0:5] == "go to":
                    os.system(["wmctrl", "-a", pinger[7:]])
                elif pinger[0:9] == "workspace":
                    pinger = pinger[10:len(pinger)]
                    if pinger == 'one':
                        pinger = 1
                    num = str(int(pinger) - 1)
                    os.system(["wmctrl", "-s", num])
                elif pinger[0:4] == "open":
                    pinger = pinger[5:]
                    os.system(["nohup", pinger])
```

Could this code also be implemented as 3 plugins?

**Comment by mmkumr** [comment]
Actually It is better to implement these 3 as single plugin. And actually I will also add some comments to this file. I forgot to do that sorry for that.

**Comment by mmkumr** [comment]
Soundcard is a beta software. It is having some bugs. When I am trying to implement it. I am encountering an error 'abort (core dumped)' most of the time.

File: jarviscli/plugins/stt.py [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+            else:
+                print(pinger)
+                if listen:
+                    line = pinger
+                    line = _jarvis.precmd(line)
+                    stop = _jarvis.onecmd(line)
+                    stop = _jarvis.postcmd(stop, line)
+
+        except LookupError:
+            engine.say('Audio cannot be read!')
```
pnhofmann: Any reason ``_jarvis.speech.text_to_speech`` isn't user here?

pnhofmann:
> Soundcard is a beta software. It is having some bugs. When I am trying to implement it. I am encountering an error 'abort (core dumped)' most of the time.

Worked better when I tried it yesterday - thank's for trying anyway. If it doesn't work reliable there is really no use.

mmkumr:
Can u send me your code.

pnhofmann:
Guess it would be an idea to just re-implement the Microphone-class. Currently I'm only "echo"-ing back everything the microphone records.
Main problem is to set SAMPLE_WIDTH, SAMPLE_RATE and CHUCK to correct values - didn't do this yet (but also didn't invested any time).

```
diff --git a/jarviscli/packages/record.py b/jarviscli/packages/record.py
new file mode 100644
index 0000000..b26cb5f
--- /dev/null
+++ b/jarviscli/packages/record.py
@@ -0,0 +1,50 @@
+import soundcard
+from speech_recognition import AudioSource
+import numpy
+
+
+class SoundCard(AudioSource):
+    def __init__(self, device_index=None, sample_rate=None, chunk_size=1024):
+        self._device_index = device_index
+        self.stream = None
+
+        self.SAMPLE_WIDTH = 1 # size of each sample
+        self.SAMPLE_RATE = sample_rate  # sampling rate in Hertz
+        self.CHUNK = chunk_size # number of frames stored in each buffer
+
+        self.SAMPLE_RATE = 48000
+        self.CHUCK = 4800
+
+
+    @staticmethod
+    def list_microphone_names():
+        return [microphone.name for microphone in soundcard.all_microphones()]
+
+    @staticmethod
+    def list_working_microphones():
+        return SoundCard.list_microphone_names()
+
+    def __enter__(self):
+        assert self.stream is None, "This audio source is already inside a context manager"
+        if self._device_index is None:
+            microphone = soundcard.default_microphone()
+        else:
+            microphone = soundcard.get_microphone(self._device_index)
+        self.stream = SoundCard.MicrophoneStream(microphone, self.SAMPLE_RATE)
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.stream = None
+
+    class MicrophoneStream(object):
+        def __init__(self, microphone, samplerate):
+            self._microphone = microphone
+            self._samplerate = samplerate
+
+        def read(self, size):
+            data = self._microphone.record(samplerate=self._samplerate, numframes=48000)
+            soundcard.default_speaker().play(data/numpy.max(data), samplerate=48000)
+            return data
+
+        def close(self):
+            pass
diff --git a/jarviscli/plugins/stt.py b/jarviscli/plugins/stt.py
index 074f6f3..989640c 100644
--- a/jarviscli/plugins/stt.py
+++ b/jarviscli/plugins/stt.py
@@ -1,6 +1,7 @@
 import speech_recognition as sr
 import pyttsx3
 import os
+from packages.record import SoundCard
 from plugin import plugin
 @plugin()
 def hear(jarvis, s):
@@ -11,12 +12,13 @@ def hear(jarvis, s):
     _jarvis.speech.text_to_speech("Say listen to start voice mode")
     while listen is False:
         try:
-            with sr.Microphone() as source:
+            with SoundCard() as source:
                 os.system('reset')  # for clearing the terminal.
                 print("Say listen to start listening")
                 r.adjust_for_ambient_noise(source)  # Eleminating the noise.
                 audio = r.listen(source)  # Storing audio.
                 pinger = r.recognize_google(audio)  # Converting speech to text
+                print(pinger.lower())
             try:
                 if (pinger.lower() == "listen"):
                     listen = True
@@ -32,7 +34,7 @@ def hear(jarvis, s):
     while listen is True:
         print("Say somthing")
         try:
-            with sr.Microphone() as source:
+            with SoundCard() as source:
                 r.adjust_for_ambient_noise(source)
                 audio = r.listen(source)
                 pinger = r.recognize_google(audio).lower()
diff --git a/requirements.txt b/requirements.txt
index faea4bf..bf5a25a 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,3 +26,4 @@ https://github.com/appi147/pyaiml3/archive/master.zip ; python_version >= '3.5'
 speedtest-cli
 SpeechRecognition
 pyaudio --global-option='build_ext' --global-option='-I/usr/local/include' --global-option='-L/usr/local/lib'
+soundcard
```

pnhofmann:
Soundcard isn't such imported at this moment. I would propose, we could keep soundcard in mind, but stay with portaudio for now. We can add another backend any time.

I've pushed patches to handle errors during installation of pyaudio. If pyaudio can't be installed, voice recognition will be disabled and rest of Jarvis will work just like it did before voice control.

I'm perfectely happy how this PR is right now. Please review and tell me, if you are OK with my changes and if this can be merged.

mmkumr:
Now, `pyaudio` will be installed automatically by recognizing the distribution and os type.

mmkumr:
Checkout the changes. If everything works fine then merge the branch.:)

pnhofmann:
There is a reason I typically avoid to install a packages on multiple linux-distributions... It just creates so much trouble...

* On [opensuse](https://software.opensuse.org/package/python3-PyAudio?search_term=python3-PyAudio) this package is called python3-PyAudio. (so PyAudio with capital letters)
* On [Arch](https://www.archlinux.org/packages/?sort=&q=pyaudio&maintainer=&flagged=) the the python2 version has the number - so it is 'python2-pyaudio' and 'python-pyaudio'
* On [Fedora,](https://apps.fedoraproject.org/packages/s/portaudio) it is 'python2-pyaudio' and 'python3-pyaudio'
* On [Gentoo](https://packages.gentoo.org/packages/dev-python/pyaudio) it is just called pyaudio
 
Also ``distro.os_release_attr('id_like')`` on my system returns ``'suse opensuse'`` not ``'suse'`` (I'm using Leap 15.0).

So basically the usual stuff when dealing with different distributions ;). Updates the install script.

mmkumr:
Ok I will reset head to previous commit.

pnhofmann:
> Ok I will reset head to previous commit.

Please don't! Should probably have made clearer: I'm fine with the installpyaudio.py script. Just needed bit tweaking - and already pushed an improved version.

mmkumr:
Sorry I have seen your reply after pushing.:(

pnhofmann:
> Sorry I have seen your reply after pushing.:(

No problem - was my fault not explaining what I'm doing ;).

mmkumr:
It's ok. Does everything works fine.

pnhofmann:
Year, so big thanks for creating this one!

PR #409, opened 2019-02-03
Added battery plugin

shahlin:
Provides information about battery such as the percentage, time remaining before empty, vendor, technology and energy consumption.

### DIFF
--- a/jarviscli/plugins/battery.py
+++ b/jarviscli/plugins/battery.py
@@ -1,10 +1,10 @@
 import subprocess
-from plugin import plugin
+from plugin import plugin, LINUX
 
 VALID_OPTIONS = ['status', 'vendor', 'energy', 'technology', 'remaining']
 
 
-@plugin()
+@plugin(plattform=LINUX, native='upower')
 def battery(jarvis, s):
     """
     Provides battery status like battery percentage
--- a/jarviscli/plugins/battery.py
+++ b/jarviscli/plugins/battery.py
@@ -1,0 +1,75 @@
+import subprocess
+from plugin import plugin
+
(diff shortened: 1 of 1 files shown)


pnhofmann:
Thanks!

``upower`` is something linux specific or is there a mac version available? Which of course would be no problem - not every plugin needs to support every platform.
Howerever does mean, we should only enable this feature if we are on a linux-machine with upower installed. Since how to do that is not well documented (work in progess), I have added 2-line commit to do so. Hope you are OK.

I just don't know: Does battery always have this path: '/org/freedesktop/UPower/devices/battery_BAT0'? On my FreeBSD machine it's available as '/org/freedesktop/UPower/devices/battery_devxbatteryx0' - I'm honestly not the freedesktop-expert.

I'll test this plugin on my linux system, investigate if there is another way than hard-coded '/org/freedesktop/UPower/devices/battery_BAT0' and merge afterwards.

shahlin:
Yes, my bad, I should have made it Linux platform only. I tried it with Mac and it doesn't work.
I tried testing on other Linux distros and it seemed to work on all for now. Here are the screenshots:

- [Ubuntu](https://i.imgur.com/LEP95am.png)
- [Debian](https://i.imgur.com/ggefl5x.png)
- [Fedora](https://i.imgur.com/i0g03uA.jpg)
- [Manjaro](https://i.imgur.com/qfgbmsh.png)
- [Antergos](https://i.imgur.com/B2FO0t0.png)

pnhofmann:
Finally had time to test... Works really nice! Interessting, FreeBSD called battery_BAT0 different - thanks for testing different different Linux-Distributions; looks really like on Linux it's allways BAT0.

Actually really like this one! Thanks!

PR #429, opened 2019-02-28
do cleanup

appi147:
- remove _config.yml and NEW_FUNCTIONALITIES.md
- remove aiml related stuff


appi147:
I have a question.
Since, cricket is broken shall I rename cricket.py as cricket.py.broken or something until `pycricbuzz` is fixed or the package is fixed.

Issue #428, opened 2019-02-27
Cricket is broken


appi147:
and pycricbuzz is also broken.
Cricbuzz has changed their contents of webpage which used to be scraped.

Falling back to previous versions doesn't help either.

appi147:
[https://github.com/codophobia/pycricbuzz/commit/ad29cedf88d782ac9596542096476403ac3b66b1](https://github.com/codophobia/pycricbuzz/commit/ad29cedf88d782ac9596542096476403ac3b66b1)

PR #431, opened 2019-02-28
fix cricket package

appi147:
So, I have tried to fix cricket.py. Should work atm.
But if some error occurs I am willing to maintain it.


appi147:
So I noticed that I have used print() instead of jarvis.say() in diff. Could it cause issues?

appi147:
closes #428

appi147:
Well Jarvis can't speak scorecards or so long commentaries. By mistake done the right thing

Issue #432, opened 2019-02-28
public ip v6 address


skerrepy:
on public ip v6 address:
curl: (7) couldn't connect to server

pnhofmann:
You probably don't have a (public) ipv6 address... This message should definitely be improved to something more meaningful like 'not available'.
Thanks for reporting!

Issue #425, opened 2019-02-26
Curvesketch have an error...


Zml99:
I tried to open the curvesketch and it give this error:
```
Traceback (most recent call last):
  File "/home/nicolai/Jarvis/jarviscli/CmdInterpreter.py", line 143, in try_do
    do(s)
  File "/home/nicolai/Jarvis/jarviscli/plugin.py", line 116, in __run_method
    run_method(jarvis, s)
  File "/home/nicolai/Jarvis/jarviscli/plugins/evaluator.py", line 266, in curvesketch
    term = solve_y(term)
  File "/home/nicolai/Jarvis/jarviscli/plugins/evaluator.py", line 224, in solve_y
    return solve_y("({}) -y".format(s))
  File "/home/nicolai/Jarvis/jarviscli/plugins/evaluator.py", line 216, in solve_y
    results = sympy.solve(s, y)
  File "/home/nicolai/Jarvis/env/lib/python3.6/site-packages/sympy/solvers/solvers.py", line 1162, in solve
    solution = _solve(f[0], *symbols, **flags)
  File "/home/nicolai/Jarvis/env/lib/python3.6/site-packages/sympy/solvers/solvers.py", line 1498, in _solve
    f_num, sol = solve_linear(f, symbols=symbols)
  File "/home/nicolai/Jarvis/env/lib/python3.6/site-packages/sympy/solvers/solvers.py", line 2073, in solve_linear
    n, d = eq.as_numer_denom()
  File "/home/nicolai/Jarvis/env/lib/python3.6/site-packages/sympy/core/add.py", line 446, in as_numer_denom
    content, expr = self.primitive()
  File "/home/nicolai/Jarvis/env/lib/python3.6/site-packages/sympy/core/add.py", line 916, in primitive
    c, m = a.as_coeff_Mul()
AttributeError: 'Tuple' object has no attribute 'as_coeff_Mul'

```

pnhofmann:
Thanks for reporting this one! Curvesketch needs an argument (e.g. curve sketch x**2+10x-5).

I'll add a meaningful message later.

Zml99:
Ohh....I didn't knew that. I will try that way next time.

Issue #418, opened 2019-02-16
Show me pics of.... Command not working.


jsgtihub:
When entering 

show me pics of " "

I get 

I could not identify your command...

jadsonlucio:
I believe the command you are looking for is: display me pics of.
But it would be interesting if there was a alias system that associated display and show.

jadsonlucio:
I believe the command you're looking for is: display pics of.
But it would be interesting if there was a alias system that associated display and show.

YasmineBOU:
Hi!

We're students of the University of Paris 8 in France.
In one of our subjects, we have to contribute in an open source project.

We would like to know if we could work on it?

pnhofmann:
@topic
``But it would be interesting if there was a alias system that associated display and show.``
Luckily there is in fact an alias-system ;). Yes, should be no problem aliasing "show" to "display"!  

@YasmineBOU 
Please **always** open a new issue for a new topic. I can't really move your comment. So to answer your question anyway:
This is basically a first time contributors project for beginners to learn. This project is designed to be easy. Just saying ;).
If this is OK for your course (and there are really no requirements that you have to contribute to a bigger project) - of course.

**Any contributions are welcome!** Just open an issue and explain, what you want to implement. Or look at the opened issues, if you want to fix something described there. Or just open an issue "Request for contribution" and we will find something for you to do ;).

Issue #415, opened 2019-02-16
Music Playback Issue


jsgtihub:
When entering the input music I get the following message.
Song name doesn't exist. ( play 'song name' )

pnhofmann:
This message is somehow misleading.

You didn't specify a song name - e.g. ``music wonderful tonight``.  Therefore the song name does not exist - not because of a error but just because none was specified ;). Should probably be updated to something like: "You didn't specify a song name" or "Please add a song name" to avoid confusion.

PR #433, opened 2019-02-28
Various Bug fixes

pnhofmann:
* Fix curvesketch crash with no argument #425
* Add alias 'show pics' for 'display pics' #418
* Update misleading error message in 'music' #415
* Add meaningful message if no ipv6 available #432
* Fix relative path issue with nltk

### DIFF
--- a/jarviscli/plugins/converted.py
+++ b/jarviscli/plugins/converted.py
@@ -97,7 +97,7 @@ def do_update(self, s):
     location: Updates location.
     system: Updates system.
     """
-    self = self_jarvis
+    self = self._jarvis
 
     if "location" in s:
         location = self.memory.get_data('city')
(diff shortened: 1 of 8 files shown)


jarviscli/plugins/converted.py
```diff
@@ -97,7 +97,7 @@ def do_update(self, s):
     location: Updates location.
     system: Updates system.
     """
-    self = self_jarvis
+    self = self._jarvis
```
appi147: I was going to point this out along with  some other errors.
In line 105 and 106 print_say() is undefined
My  editor shos ConnectionError is undefined as well as CONNECTION_ERROR_MSG

pnhofmann:
Thanks @appi147 ! Should really have tested this jarviscli/plugins/converted.py module better.

PR #436, opened 2019-03-17
fix ErrorValue in 'score' function of cricket plugin

Farlgerann:
catch ValueError to avoid crash of cricket plugin


pnhofmann:
Works perfectly; Thanks!

PR #444, opened 2019-03-25
Adding an explanation about the CURRENCY CONVERTER output.

JonanOribe:
Now its gives you something like: "100.0 AUD are equal to 62.7700 EUR" as response, instead of only "62.7700"

### DIFF
--- a/jarviscli/plugins/currencyconv.py
+++ b/jarviscli/plugins/currencyconv.py
@@ -46,8 +46,9 @@ class Currencyconv():
             result = b.convert_btc_to_cur(Decimal(amount), to)
         else:
             result = c.convert(fr, to, Decimal(amount))
-
-        jarvis.say(str(amount)+" "+fr+" are equal to "+str(result)+" "+to)
+        
+        outputText=str(amount)+" "+fr+" are equal to "+str(result)+" "+to
+        jarvis.say(outputText)
 
     def find_currencies(self):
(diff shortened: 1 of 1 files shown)


appi147:
Just give spaces around arithmetic operators and test will pass

jarviscli/plugins/currencyconv.py
```diff
@@ -46,8 +46,9 @@ def currencyconv(self, jarvis, amount, fr, to):
             result = b.convert_btc_to_cur(Decimal(amount), to)
         else:
             result = c.convert(fr, to, Decimal(amount))
-
-        jarvis.say(str(result))
+        
+        outputText=str(amount)+" "+fr+" are equal to "+str(result)+" "+to
```
appi147:
```suggestion
        outputText = str(amount) + " " + fr + " are equal to " + str(result) + " " + to
```

jarviscli/plugins/currencyconv.py
```diff
@@ -46,8 +46,9 @@ def currencyconv(self, jarvis, amount, fr, to):
             result = b.convert_btc_to_cur(Decimal(amount), to)
         else:
             result = c.convert(fr, to, Decimal(amount))
-
-        jarvis.say(str(result))
+
```
appi147:
```suggestion

```

Issue #441, opened 2019-03-25
Movie: two methods are broken


appi147:
Movie search and movie info is broken because of incorrect method defined in line 220.
Variable `move_attributes` is iterated however it is undefined in line 228 of movie.py.
Error log:
```
movie info interstellar

Some error occurred, please open an issue on github!
Here is error:

Traceback (most recent call last):
  File "jarviscli/CmdInterpreter.py", line 151, in try_do
    do(self, s)
  File "jarviscli/plugin.py", line 209, in run
    self.get_plugins(command).run(jarvis, new_s)
  File "jarviscli/plugin.py", line 219, in _plugin_run_with_network_error
    run_func(jarvis, s)
  File "jarviscli/plugin.py", line 201, in run
    self._backend[0](jarvis.get_api(), s)
  File "/home/appi/Public/github/Jarvis/jarviscli/plugins/movie.py", line 150, in movie_info
    get_movie_info(jarvis, data)
  File "/home/appi/Public/github/Jarvis/jarviscli/plugins/movie.py", line 228, in get_movie_info
    for attribute in movie_attributes:
NameError: name 'movie_attributes' is not defined

```

```
movie search interstellar

ID   Movie title
1    Interstellar (2014)
2    Interstelar 2: Operation Terra 2040 (2016)
3    Interstellar Civil War: Shadows of the Empire (2017)
4    Interstellar Wars (2016)
5    Interstelar (2014)

Please enter ID to know more(q - quit):
1

Some error occurred, please open an issue on github!
Here is error:

Traceback (most recent call last):
  File "jarviscli/CmdInterpreter.py", line 151, in try_do
    do(self, s)
  File "jarviscli/plugin.py", line 209, in run
    self.get_plugins(command).run(jarvis, new_s)
  File "jarviscli/plugin.py", line 219, in _plugin_run_with_network_error
    run_func(jarvis, s)
  File "jarviscli/plugin.py", line 201, in run
    self._backend[0](jarvis.get_api(), s)
  File "/home/appi/Public/github/Jarvis/jarviscli/plugins/movie.py", line 208, in movie_search
    get_movie_info(jarvis, data)
  File "/home/appi/Public/github/Jarvis/jarviscli/plugins/movie.py", line 228, in get_movie_info
    for attribute in movie_attributes:
NameError: name 'movie_attributes' is not defined


```

qwireq:
Can you assign this issue to me?

appi147:
Sure go ahead

[...]

PR #447, opened 2019-03-26
Solution for two broken methods in Movie.py

qwireq:
The link to the issue: #441. 
The errors were appearing because of the empty `movie_attributes` variable in `get_movie_info` function from `Movie.py` file.  I acted according to the text given on 222 line of code, so `movie_attributes` variable was initialized from the user's input. If entered attributes match with the information retrieved from IMDB using ImdbPy, the data is displayed. If no match is found, `no information retrieved` placeholder is displayed. Finally, if the user does not enter any attribute, then only the URL of the movie is displayed.
This is my first PR. Let me know if you have any feedback!

### DIFF
--- a/jarviscli/plugins/movie.py
+++ b/jarviscli/plugins/movie.py
@@ -222,12 +222,16 @@ def get_movie_info(jarvis, data):
     Takes a movie attributes as input and prints them accordingly
     """
     global movie_attributes
-
+    
     jarvis.say('')
-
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/movie.py
```diff
@@ -223,11 +223,16 @@ def get_movie_info(jarvis, data):
     """
     global movie_attributes
```
appi147: Is this line required?

qwireq: I don't think it is required, because the variable is not used in any other functions. I left it unchanged because it was global before I started working on this issue. I am going to change it and do a commit.

PR #449, opened 2019-03-26
Write agenda plugin v0.1

JonanOribe:
A little script that allows the user to write key:value data in a CSV. This is the entry point for a Read_agenda plugin that recovers these data and future improves like a  reminder of events ,bearing in mind the dead line that they will have. For the next version Write_agenda will allow to add new lines to your agenda, at this moment every time you execute the script it will overwrite the old agenda with the new data

### DIFF
--- a/jarviscli/plugins/write_agenda.py
+++ b/jarviscli/plugins/write_agenda.py
@@ -2,8 +2,8 @@ from plugin import plugin
 import csv
 
 
-@plugin("write_agenda")
-def agenda(jarvis, s):
+@plugin("write agenda")
+def write_agenda(jarvis, s):
     exit = True
     csv_columns = ['Title', 'Description']
     mydict = {}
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/write_agenda.py
```diff
@@ -0,0 +1,30 @@
+from plugin import plugin
+import csv
+
+
+@plugin("write_agenda")
```
appi147:
```suggestion
@plugin("write agenda")
```

JonanOribe: Really?? I hate Travis jajaj.Thanks for the help!!

JonanOribe:
I think that is something about the way I make the "While" at line 11

appi147:
Add read agenda in this file, from PR #451

appi147:
Yes that is a problem, but we follow this naming

PR #450, opened 2019-03-27
Random password generator

JonanOribe:
Auto generate a fixed length password

### DIFF
--- a/jarviscli/plugins/random_password.py
+++ b/jarviscli/plugins/random_password.py
@@ -5,7 +5,7 @@ import string
 
 @plugin("random password")
 def random_password(jarvis, s):
-    stringFail=True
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/random_password.py
```diff
@@ -0,0 +1,13 @@
+from plugin import plugin
+import random
+import string
+
+
+@plugin("random_password")
```
appi147:
```suggestion
@plugin("random password")
```

appi147:
Can also take an input for length of password

JonanOribe:
> Can also take an input for length of password

Great idea, when I will have a moment, I'm going to work around that point

PR #451, opened 2019-03-27
Read agenda plugin that complements Write Agenda

JonanOribe:
Just a method for easy reading agenda.csv


appi147:
@JonanOribe Can you move this code to write_agenda PR. Just keep all agenda codes(read agenda and write agenda) in a single file. No need to create two files for same plugins. 
Also `write agenda` and `read agenda` is better than `write_agenda` and `read_agenda`

PR #465, opened 2019-04-07
New Plugin for Assembly Language Decoder

usman-kakakhel:
Hi,
I noticed that jarvis had some educational purpose plugins such as plot, limits, evaluate, equations. I thought maybe jarvis it would be useful if I add an Assembly(mips) language decoder which gives the machine code of assembly commands and gives Assembly commands for machine code.

I have already run the flake 8 check so I hope this feature would be accepted.
Thankyou


usman-kakakhel:
Could you guys please check the pull request and accept it? Or if there is any problem with it, can you please tell me so that I can fix it?

pnhofmann:
Mips - quite uncommon today, isn't it? Was just checking Wikipedia ;). But of course you are right: Some bit educational purpose is surely nothing we would be disaffected about.

If I understand it correctly, this is for early 32 bit MIPS I? Should be specified somewhere. 

The 'coms.txt' file (should be renamed to something like 'mips_coms.txt') - just wondering, you don't have by chance a nice MIPS-specification we could link for anyone who is interested (including me)? Did you put these instructions in this format or can this file be found somewhere?

Code is perfectly readable.

Just noticed - ``__hexToBin`` and ``__binToHex``. Did you know, Python has actually some pretty sweat Build-Ins for these jobs.
``int(str, base)`` converts str to an integer - and with ``format`` an integer can be converted to binary or hex. So, this could be simply written as:

```
__hexToBin(self, myHex):
    return "{:b}".format(int(myHex, 16))

__BinToHex(self, myBin):
    return "{:X}".format(int(myBin, 2))
```

Will check some details of your code latterly. But overall really like this plugin. Good job!

usman-kakakhel:
Thankyou @pnhofmann  for your response, it indeed is uncommon but it is used for educational purposes even now. Also this is not just for MIPS 1 but rather it has several mips 2 instructions such as TNEI, TGEIU. Furthermore it also has instructions ranging upto mips 4 such as SC, MOVN, MOVZ. This file named coms.txt was made by myself and the 70 basic mips instructions were decoded and the pattern of each was done by hand. Not every MIPS 1 to Mips 4 instruction is available I think but most of them are there. The ones which are widely used. Yeah this is just for 32 bit instructions. This file was made by me so you wont find it in this pattern online and it might be hard to find compilation any compilation online. I will indeed change the name of the file to mips_coms.txt. To the issue of using python conversions, I was doing that before but I was getting some technical errors. The time it would have taken me to fix those would have been more than the time it took me to make these two small application specific methods. So if it is indeed necessary to change these two custom methods with python ones, I would prefer it this way. It would be nice if we didnt change them. But if it is required, I would change it of course. For mips specs, you can learn it by looking at the book Digital design and computer Architecture second edition by David Money Harris http://hron.fei.tuke.sk/~adam/cas/Digital%20Design%20and%20Computer%20Architecture%20(2nd%20Ed).pdf.

 There is a MIPS IDE MARS which you can easily google and download. In the help section of the  IDE you can easily find explanations to some instructions. http://courses.missouristate.edu/KenVollmar/MARS/

(https://inst.eecs.berkeley.edu/~cs61c/resources/MIPS_Green_Sheet.pdf) mips green card has opcodes and the types of instructions mips has and the way they have to be converted.

jarviscli/plugins/MipsConverter.py
```diff
@@ -0,0 +1,534 @@
...
+        in this case ADDI. There should be a space between the
+        instruction keyword and the first register and then the
+        rest can be separated with commas.
+
+    A good practice would be to give a space between any separate part of
+    the instruction such as between keyword and register, or keyword
+    and immediate, or register and immediate.
+    """
+    def __call__(self, jarvis, s):
+        self.__init__()
```
pnhofmann: Almost forgot: This line is obsolete, because __init__() is called on class construction and does not need be called manually.

> So if it is indeed necessary to change these two custom methods with python ones, I would prefer it this way. 

Ok, then we'll keep the methods like you prefer.

usman-kakakhel:
Firstly, I looked at the test results that you gave me. The the Hex to assembly conversion was good but the assembly to Hex conversion was wrong. There is a reason for that. If you look at the test data below:
0399e021   - addu gp gp t9          ---   0x0010          - addu $gp $gp $t9 
from 03399e021 the expected assembly is addu gp gp t9 and we do indeed get addu $gp $gp $t9  but
since the hex is 0399e021, the hex printed by my code from addu gp gp t9 should have been 0399e021
while it is not. The main reason for this is that whenever you are writing machine code, you always write $ before a register. The syntax for assembly is exactly this. Now if you do not write a $ before a register, you cannot differentiate between a decimal immediate and a register value. The reason why this test data does not have any $ before its commands is that they support only register names such as $zero while my program runs with register name and code meaning $zero and $0 are the same. To conclude, the test file you provided, gives results for hex to assembly with a very small problem(which I fixed) but assembly to hex does not look good as the input is not in the expected format. And I cannot throw exceptions for every command that has a wrong syntax because not having a $ in an assembly does not necessarily mean that the instruction is wrong.

The instructions that are not supported are pseudo instructions while my code supports only basic mips instructions. Pseudo instructions usually vary from architecture to architecture so we cannot make a generalization for them. 

No need to worry about outputs, as I have already tested manually with a lot of test data and the results are promising. 

Removed the init() as required.

As for the unit test, I may be able to make an assembly program with results for testing but I need to know how long the program should be. How many instructions would you want. Also I looked over at the way how to make a unit test and looked at unit tests from other plugins but unfortunately could not not figure out how to make a unit test. So I would provide you with the assembly program and its expected result with as many instructions as you require and you might have to help me make the test.

In the mean time, I am ready for merge now as the plugin is has been tested by me with a lot of test data.

pnhofmann:
Understood!

Unit test is not required - honestly it's actually something on my personal "To-Improve"-list ;). I've started writing [documentation](https://github.com/sukeesh/Jarvis/blob/master/doc/TESTING.md) how to best test Jarvis-code, but could really be improved.

So - I'll just merge.

Thanks for your great work!

PR #484, opened 2019-04-29
Plugin to open website for jarvis

rogervieiraa:
Basically, this will work in 2 ways.

### Input a complete link:
`> open website www.google.com`
This one will validate if the link is valid before, after that he open.
Until now the complete link only works if ends with 'com', none of others websites terminations is working by now.

### Use a dictionary with popular links (create a csv file to save the files)
`> open website github username`
This will open the respective GitHub page.

If the name isn't on dictionary and isn't a validated link the plugin will print a message, that cannot open the website.

I have some question about the String input in the plugins:
- Why always come without the dots?
- Have some way to when identifying this plugin the dots stay on the input?

### DIFF
--- a/jarviscli/plugins/OpenWebsite.py
+++ b/jarviscli/plugins/OpenWebsite.py
@@ -33,7 +33,7 @@ class OpenWebsite:
         elif self.verify_link():
             webbrowser.open(self.main_link)
         else:
-            print("Sorry, I can't open this link, fix it and try again")
+            print("Sorry, I can't open this link.")
 
     def has_on_saved_links(self):
         websites_csv = \
--- a/jarviscli/plugins/OpenWebsite.py
+++ b/jarviscli/plugins/OpenWebsite.py
@@ -24,12 +24,12 @@ class OpenWebsite:
     def __call__(self, jarvis, link):
         inputs = link.split(' ', 1)
         self.main_link = inputs[0]
-        complement = ""
+        self.complement = False
         if len(inputs) > 1:
-            complement = inputs[1]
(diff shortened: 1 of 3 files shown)


appi147:
Opening username is not consistent.
1: Opening medium without username leads to 404
2: Opening reddit with leads to subreddit (u/)

rogervieiraa:
I think now is fixed

rogervieiraa:
@appi147

appi147:
Please review this. I am super busy.

jarviscli/plugins/OpenWebsite.py
```diff
@@ -0,0 +1,87 @@
...
+                    self.main_link = information[1]
+                return True
+
+        return False
+
+    def verify_link(self):
+        if ((self.main_link[:8] != "https://" and
+             self.main_link[:7] != "http://" and
+             self.main_link[:3] != "www") or
+                ("com" not in self.main_link)):
```
pnhofmann: There are lots of other top-level domains besides '.com'...

https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains

Instead of this check I would just resolve the domain. With Python e.g.:

```
socket.getaddrinfo("example.org", 80)
```
https://docs.python.org/3/library/socket.html#socket.getaddrinfo

And if this does not throw a socket.gaierror domain should be ok ;).

This module is also always usfull when dealing with urls:

https://docs.python.org/3/library/urllib.parse.html

pnhofmann:
Hi! Made some changes, please review and tell me if you are OK with them.

PR #489, opened 2019-04-30
Converters for Time and Distance

skpanagiotis:
I have made a length unit converter and a time unit converter.

The length unit converter supports nanometer, micrometer, millimeter, centimeter, decimetre, meter, kilometer, mile, yard, foot, inch.
The user can insert an amount, then the length measurement unit of the amount and the length measurement unit the user want to convert the amount.
For example:
`Enter an amount: ` 2
`Enter from which unit: ` km
`Enter to which unit: ` m
And the plugin will print:
`2.0 kilometers is equal to 2000.0 meters`

The time unit converter supports picosecond, nanosecond, microsecond, millisecond, second, minute, hour, day, week, month, year.
The user can insert an amount, then the time measurement unit of the amount and the time measurement unit the user want to convert the amount.
For example:
`Enter an amount: ` 2
`Enter from which unit: ` wk
`Enter to which unit: ` d
And the plugin will print:
`2.0 weeks is equal to 14.0 days`

Because microsecond or micrometer use in their symbols (μs, μm) the special character `μ`, when you want to type microsecond you need to type `mus` and when you want to type micrometer you need to type `mum`.

Also if a number has many decimal places, I gave the opportunity to the user to add the precision he/she want, with maximum 12 decimal places.

I also add unit tests for the two plugins.


StefanosChaliasos:
Please provide a more detailed description of your contribution (features, examples, etc.). Your code should also pass flake linter, and you should add some unit tests for your plugins.

StefanosChaliasos:
Nice work!

PR #505, opened 2019-06-27
Added buy plugin

ApostPap:
Search a term in Amazon or Ebay.

### DIFF
--- a/jarviscli/plugins/buy.py
+++ b/jarviscli/plugins/buy.py
@@ -38,7 +38,6 @@ def buy(jarvis, s):
             return None
 
     webbrowser.open(endURL)
-    #openBrowser(jarvis, endURL)
 
 
 # Check is shop is supported and creates the url for searching on that shop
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/buy.py
```diff
@@ -0,0 +1,91 @@
...
+def oneLineCommand(shop_input, search_term):
+    endURL = ""
+    startingURL = shop(shop_input)
+    if len(startingURL) > 0:
+        endURL = generateURL(startingURL, search_term, True)
+    return endURL
+
+
+# Opens Browser, new tab with the search url
+def openBrowser(jarvis, endURL):
```
pnhofmann: Looks good!

Just noticed: Python in fact has a module 'webbrowser'.

https://docs.python.org/3.5/library/webbrowser.html#webbrowser.open

So you can just write ``webbrowser.open(url)``. We already use webbrowser e.g. in news or picshow plugin.

Otherwise I have nothing to complain ;) !

ApostPap: Ok I switched to webbrowser. Thanks for your input.

pnhofmann:
Thanks!

Issue #487, opened 2019-04-29
Translate command should be able to understand the full request in one line


ceduardodfernandes:
It would be much better if the translate command could be understood by Jarvis in one single line as well:

"translate english pt what are you doing?"

Matthew237:
I am now currently working on this. I have it working provisionally and I will create a pull request once I am sure the code meets the PEP-8 requirements and after I have added some better natural language processing for translation (right now you have to enter the command somewhat robotically for it to work).

pnhofmann:
Thanks @Matthew237 
#517

Issue #486, opened 2019-04-29
Translate command don't recognize spaced language


ceduardodfernandes:
The command translate do not understand if the language contains white spaces at begin or end of the request

Matthew237:
I have just created a pull request that fixes this issue

pnhofmann:
#516

PR #517, opened 2019-07-19
Improvements to translate.py and small typo fix

Matthew237:
There is more detailed information in the commit messages, but this basically adds the ability to type translation requests in one line, which fixes issue #487. It also fixes a small typo in every prompt but the first where an 'I' wasn't capitalized.

Also, just in case you didn't see the comments on my last pull request, that pull fixed issue #486 so you can close that issue too if you want :)

Sorry for the mess of commits in this pull request, I'm a complete beginner when it comes to Github so I'm still getting used to how this all works.

There are still some improvements to be made here, but this allows you to ask for translations in a pretty natural way. I think to get it any better I would have to use some proper NLP but I'm not sure how to do that. I'm currently reading about how to use NLTK and the likes so if I start to understand it I might try to improve this, and Jarvis in general, further.

Finally, I did run the test.sh script on this code so it should follow all PEP-8 guidelines, but just let me know if it doesn't and I'll fix it up :)


pnhofmann:
Hmm, throws an error for the example in #486.
```
~> What can I do for you?
translate english pt what are you doing?
Some error occurred, please open an issue on github!
Here is error:

  File "/home/phil/git/Jarvis/jarviscli/CmdInterpreter.py", line 188, in try_do
    do(self, s)
  File "/home/phil/git/Jarvis/jarviscli/plugin.py", line 226, in _plugin_run_with_network_error
    run_func(jarvis, s)
  File "/home/phil/git/Jarvis/jarviscli/plugin.py", line 208, in run
  File "/home/phil/git/Jarvis/jarviscli/plugins/translate.py", line 27, in translate
    elif srcs in SPECIAL_CASES:
UnboundLocalError: local variable 'srcs' referenced before assignment
```

Plus please add an example to docstring.

> Sorry for the mess of commits in this pull request, I'm a complete beginner when it comes to Github so I'm still getting used to how this all works.

No problem at all! We are here for learning ;).

By the way, in case you didn't knew: If you want to read/look-up something, really good (and free!) book with really everything you need to know about git: https://git-scm.com/book/en/v2

Matthew237:
Oh, I've just had a look at my code and it looks like it's just a silly mistake. I think I've just put "srcs" in a place where I meant to put "word". I must have just not hit that part of the code during my testing. I'll get it fixed when I have a minute.

Also, sorry if it is a silly question but what do you mean by putting an example in the docstring?

appi147:
It says for me to install punkt for nltk. After installing, it does not work for me, it goes to default method. For docstring, [read here](https://www.datacamp.com/community/tutorials/docstrings-python). Its a way to write comments under method name with an example

Matthew237:
Yeah, I think nltk needs punkt for the word_tokenize() function that I used. I thought it was already installed for the project, so if it isn't I can try to find a new way of doing that that doesn't require extra installs.

Also, what command did you enter that went to the default method? I had to disable the use of language codes on the inline command because it causes conflicts. For example if I were to input "Translate hi from English to French" and the plugin was searching for language codes, then it would find "hi" (Hindi's language code) as the source language and would not perform as expected.

Obviously, there is still an issue if the phrase to be translated contains a language name, but I could not think of a way around this and this is a relatively niche case. Also, if your phrase did contain a language, it would still be possible to use the default method to do the translation.

appi147:
OK cool. I used language codes.

appi147:
Ok just add punkt in setup.sh

appi147:
Try this: `translate english hindi how are you`
Output:
```
[31m~> Hi, what can I do for you?
[39mtranslate english hindi how are you

[en] you
    ->
[hi] आप
[pron.] aap
```

Issue #519, opened 2019-07-20
Run commands directly from the terminal


aeciorc:
I thought it would be nice to able to get jarvis to run a command directly from the terminal, i.e, without having to initialize it and wait for the prompt.
For example:
 `jarvis roll` would initialize jarvis, run the dice.py plugin, print the ouput and exit. 

If it is approved, I would love to implement it myself.

pnhofmann:
Hi!
Like this idea ;). Looking forward to your PR!

aeciorc:
Awesome, I just submitted it :)

pnhofmann:
Thanks!
#520

PR #520, opened 2019-07-22
Let Jarvis run commands directly from the terminal  #519

aeciorc:
Calling Jarvis with a command-line argument now starts an execution mode which attempts to run that argument as a command and terminates.

setup.sh was modified to generate a `/jarvis` file which that handles command-line arguments, so the setup must be re-run in order for the new feature to work.


pnhofmann:
Nice work!

PR #523, opened 2019-07-26
Personal configuration

JonanOribe:
Save personal information about the user for future configuration


appi147:
age changes every year, maybe write a method to calculate age with current age

JonanOribe:
> age changes every year, maybe write a method to calculate age with current age

For sure, noob fail

PR #527, opened 2019-07-26
New personal info plugin

JonanOribe:
Now with "born date" instead of "age" because it will be more usefull if you want to calculate the birthday,the age or another things about the user


appi147:
Hello @JonanOribe , can you please add myinfo.csv to gitignore

JonanOribe:
@appi147  .gitignore has been update with myinfo.csv

Issue #531, opened 2019-07-27
Make require for windows


appi147:
Windows does not seem to respect @require(linux)

appi147:
@pnhofmann Any ideas

pnhofmann:
Yep, isn't implemented yet - like documented here:
https://github.com/sukeesh/Jarvis/issues/346#issuecomment-431596464
And unfortunately nothing has changed since ;(.

The @require(linux) check just wasn't build with windows in mind - basically checks if system is not macOS... Should be quite easy to fix.

appi147:
Thank you

pnhofmann:
I would also introduce require(platform=UNIX) as shortcut for Linux or MacOS.

What should we do about cygwin? Is this Windows or Linux or unix or a own platform or not supported?

appi147:
With arrival of Linux subsystem on Windows, cygwin is kinda not required to support. I am not interested in supporting cygwin. Maybe let it be unsupported.

UNIX is a good idea.

methods such as shutdown, or commands like sudo, grep, can be moved in UNIX.

appi147:
For python2.7.3+ [Possible outputs of sys.platform](https://docs.python.org/2/library/sys.html#sys.platform). Main point linux2 is used for most Linux systems.
For python3.3+ [Possible outputs of sys.platform](https://docs.python.org/3/library/sys.html#sys.platform). Main point linux is used for all linux systems

PR #532, opened 2019-07-27
add support for all platforms in plugin manager

appi147:
I tried to add all supported platforms. I don't know if UNSUPPORTED is a good idea, because we will never call @require(platform=UNSUPPORTED).
ALso, till UNIX is implemented, 
```
@require(platform=LINUX)
@require(platform=WINDOWS)
```
works fine, but 
```
@require(platform=LINUX, platform=WINDOWS)
```
gives error.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -74,20 +74,20 @@ Plugins, which are non compatible are ignored and will not be displayed.
 Current available requirements:
 
 * Network: (True, False)
-* Plattform: (plugin.LINUX, plugin.MACOS)
+* Platform: (plugin.LINUX, plugin.MACOS, plugin.WINDOWS, plugin.UNIX - equals plugin.LINUX + plugin.MACOS)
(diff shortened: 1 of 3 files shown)


appi147:
closes #531

appi147:
@pnhofmann Merge only if you are satisfied, or you are free to implement in a better way

appi147:
If this is done, most of the work will be done as there won't be error with 
@linux @macos, till perfect version is implemented for win32

pnhofmann:
> @require(platform=LINUX, platform=WINDOWS)
> gives error.

Python won't allow two parameters with same name => So implemented this Syntax:

``@require(platform=[LINUX, WINDOWS])``

pnhofmann:
> I don't know if UNSUPPORTED is a good idea, because we will never call @require(platform=UNSUPPORTED).

Think, this would be the default if no @require is specified. So don't think the UNSUPPORTED variable is required.

appi147:
Looks great

PR #533, opened 2019-07-27
Done in preparation for windows support

appi147:
basic commands


appi147:
@pnhofmann , I request to take a look at diff, if its ok.

PR #534, opened 2019-07-27
Fix Memory delete.

shreyashag:
Added tests for Memory package.


shreyashag:
Thanks for the merge!

PR #539, opened 2019-07-30
New plugin: dial code

alexandr7035:
Prints the telephone code of the country the user requested

`~> What can i do for you?`
`dial code of Belarus`
`Dial code is +375`
`~> What can i do for you?`
`phone code of United States`
`Dial code is +1`

But I'm not sure that the project really needs this feature... :)


alexandr7035:
Oh..i don't now how travis works.. 
Somthing is wrong with python2.7 support?

alexandr7035:
i've added `@require(python=PYTHON3)` and installed jarvis with python2, but there is error message on start :( Why the plugin is still loaded?

pnhofmann:
It's a syntax error.

In Python 2 print is no function - but an buildin. So python 2 'print' does not have parameters like end=''.
Synatx errors are throws the moment python read the file - before code like ``@require(python=PYTHON3)`` can be executed.

Anyway, you should better use ``jarvis.say(TEXT)``. Because that's required for stuff like text to speech to work.

Really nice and useful plugin!

alexandr7035:
> Anyway, you should better use [jarvis.say(TEXT)](https://github.com/sukeesh/Jarvis/blob/master/doc/API.md#say). Because that's required for stuff like text to speech to work.

Thanks for this link, i also noticed useful `jarvis.input()` method wich works both on python2 and python3 :)

alexandr7035:
It seems all right now

appi147:
Nice work!! @alexandr7035 . It would be great if we could also check country with dial code. If you are willing to implement that, I'll let it be open, otherwise its ready to merge.

appi147:
You can also check [this file](https://github.com/sukeesh/Jarvis/blob/master/jarviscli/utilities/lexicalSimilarity.py).

alexandr7035:
@appi147 

> Nice work!! @alexandr7035 . It would be great if we could also check country with dial code. If you are willing to implement that, I'll let it be open, otherwise its ready to merge.

I think I can continue to work on it :)

alexandr7035:
@appi147 

> You can also check [this file](https://github.com/sukeesh/Jarvis/blob/master/jarviscli/utilities/lexicalSimilarity.py).

If i can implement this, I'll open new PR later)

alexandr7035:
Thanks for merge!)

PR #544, opened 2019-07-31
what if a syatem have no voice engine, added try catch

appi147:


appi147:
saved upto 4-6 minutes on travis builds python27(darwin)

PR #550, opened 2019-08-10
Typing test that displays the user typing speed in words per minute

Sanil2108:
A new typing test feature. 

**How it works**
Just type typing test in the command.
`typingtest`

It gets a random combination of words from http://www.randomtext.me API and then as the user types, the words, it calculates WPM(words per minute) and how many letters are correct. When the time is up, or when the user has finished typing all the words, it would display the words the user typed per minute. If the WPM is >=20 and <=90, it will also display the percentage of people that are worse than the user.


jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+    new_seconds = str(new_seconds)
+
+    new_minutes = '0'*(2-len(new_minutes)) + new_minutes
+    new_seconds = '0'*(2-len(new_seconds)) + new_seconds
+
+    return new_minutes + ':' + new_seconds
+
+
+def take_input():
+    entered_letter = getch()
```
Dastar: for input you have the jarvis.input method

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+    if (entered_letter == '='):
+        game_end()
+    letter_object = info['letters'][info['position']]
+    print(letter_object.letter)
```
Dastar: ...and for printing the jarvis.say method

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+        info['time_left'] -= (time_1 - time_0)
+        if(info['time_left'] <= 0 or info['position'] == len(info['text'])):
+            game_end()
+
+
+def game_end():
+    info['state'] = GameState.ENDED
+    wpm = info['words'] / ((total_time - info['time_left']) / 60)
+    print(default_text_color+'\n\nWords per minute - {}'.format(wpm))
+    if(wpm >= 20):
```
Dastar: change it to
if wpm >= 20:

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+        with open(os.path.join(FILE_PATH, "../data/typing_test_data.csv"),
```
Dastar: what will happen if someone will delete this file?

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+                  mode='r') as file:
```
Dastar: it is not the best decision to use build-in function as a variable name.

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+            data = list(csv.reader(file))
+            for i in range(len(data)):
```
Dastar: This is the C-way to run over a list.
I see the reason why you are running over a list in this way, but maybe you can think about some better way?

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+
+def print_screen():
+    print("\033c")
+    string = 'Welcome to the typing text game. Enter the letters shown \
+        below (\'=\' to quit) - \n'
+    string += '\nScore - {}\t Time - {}\t Words -  \
+        {}\n\n'.format(info['score'], format_time(info['time_left']),
+                       info['words'])
+    for i in range(len(info['letters'])):
+        string += str(info['letters'][i])
```
Dastar: for i in info[letters]:
    string += str(i)

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+    sys.stdout.write(string)
```
Dastar: why are you using stdout.write in one place, print in another, when you have jarvis.say?

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@require(network=True)
+@plugin("typingtest")
+def typingtest(jarvis, s):
+    game_start()
```
Dastar: Can you pleeeeeeeease comment your code?

jarviscli/plugins/typingtest.py
```diff
@@ -0,0 +1,201 @@
...
+    wpm = info['words'] / ((total_time - info['time_left']) / 60)
+    print(default_text_color + '\n\nWords per minute - {}'.format(wpm))
+    if(wpm >= 20):
+        with open(os.path.join(FILE_PATH, "../data/typing_test_data.csv"),
+                  mode='r') as file:
+            data = list(csv.reader(file))
+            for i in range(len(data)):
+                if int(data[i][0]) > int(wpm) and i != 0:
+                    print('You are better than {}% of people!'
+                          .format(float(data[i - 1][1]) * 100))
```
Dastar: what will happen here if i'm the best of the best and nobody can write faster then me?

appi147:
Slow down @Dastar , he is first time contributor.

Dastar:
> Slow down @Dastar , he is first time contributor.

Yes, I understand, and it is great!
But from this stuff we are learning.

Issue #530, opened 2019-07-27
stock does not work


appi147:
API has been moved to new url, needs registration and authentication

appi147:
@pnhofmann Can we add a modifier like @require, for packages that are not supported atm, until they are fixed.

pnhofmann:
Either that or we set platform=[]

madhavkulkarni1986:
I feeel, we have two options to fix this: 

1.  Just like in the news plugin, as the user to register here https://iexcloud.io/cloud-login#/register and provide a way to add the apikey into memory. The same can be passed to the api query. There is a free tier with 500000 queries per month per key.

2. Find a new api that provides a authentication less access to stock data(https://github.com/public-apis/public-apis#finance => https://financialmodelingprep.com/ seems to be good)

Both have pros and cons. I personally would prefer to have the 2nd way, which also provides more endpoints to enhance stock plugin and include more functionality.

madhavkulkarni1986:
This issue is fixed with the new API. Please check PR #554 
Let me know if this is acceptable.

pnhofmann:
#554

PR #554, opened 2019-08-26
Modify stock plugin with new stock API

madhavkulkarni1986:
Fixing the issue: https://github.com/sukeesh/Jarvis/issues/530
Added new stock API  from https://financialmodelingprep.com/developer/docs/


pnhofmann:
Finally!
Added Attribution plus minor improvement with help command (issues the original stock plugin had, too - just noticed while reviewing this PR.

Yeah, nothing to add except big thanks for fixing our stock plugin!

PR #562, opened 2019-10-01
Added football info plugin

NilayMajorwar:
Added a new plugin for football information.
The plugin provides two options:

- Competition info: Shows the standings of the competition chosen by user
- Daily match info: Shows today's matches in the competition chosen by user

The plugin uses the football-data.org API's free tier, which means information for only tier one competitions is available.

### DIFF
--- a/jarviscli/plugins/football.py
+++ b/jarviscli/plugins/football.py
@@ -27,6 +27,8 @@ class Football():
     """
 
     def __call__(self, jarvis, s):
+        # Attribution to the football-data.org API
+        print("Football data provided by the Football-Data.org API\n")
         # Ask for user option (competition or match info)
         option = self.get_option(jarvis)
         if option is None:
@@ -163,7 +165,7 @@ class Football():
                 stList.append([team[k] for k in team])
             # Print the standings table
             print(tabulate(stList, headers=[
-                  "#", "Team", "P", "W", "D", "L", "Points", "GF", "GA", "GD"]))
+                "#", "Team", "P", "W", "D", "L", "Points", "GF", "GA", "GD"]))
             print()
 
     def matches(self, jarvis, compId):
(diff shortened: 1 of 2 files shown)


pnhofmann:
Nice addition to our cricket plugin!

Any particular reason you are mixing jarvis.say() and print()? jarvis.say() is actually better, because allows us to do some stuff like saying out load or maybe (in future) redirecting output to a GUI.  

I just never know how to handle API keys - especially in Open Source software... But if you are ok to provide your API key, that should be ok? @appi147 

Just noticed from the football-data.org FAQ:

```
Do I need to provide attribution to Football-Data API?
Yes, you'll need to include a disclaimer "Football data provided by the Football-Data.org API" in a visible section of your app/website.
```

So that's something you should probably add.

NilayMajorwar:
I am avoiding the use of `jarvis.say` for longer strings and tables. Speaking out the list of competitions takes quite some time, and the standings table when spoken out is not understandable at all.

Wouldn't it be easier if `jarvis.say` had an optional argument that mentions whether the text should be spoken out or just printed? That way, the above issues can be solved without `print` and it also maintains the Jarvis core layer between the user and the plugins.

> Do I need to provide attribution to Football-Data API?
> Yes, you'll need to include a disclaimer "Football data provided by the Football-Data.org API" in a visible section of your app/website.

Sorry, didn't notice that! I'll add the attribution to the plugin.

pnhofmann:
> Wouldn't it be easier if jarvis.say had an optional argument that mentions whether the text should be spoken out or just printed?

You’ve got a point ;). Yeah, that would be useful - I'll put it on my list.

NilayMajorwar:
I've added the attribution line to the plugin.

pnhofmann:
Thanks!

NilayMajorwar:
> optional argument that mentions whether the text should be spoken out or just printed

Can I do this one, @pnhofmann ? It seems like a simple one.

NilayMajorwar:
I've added a new issue for this topic [here](https://github.com/sukeesh/Jarvis/issues/588).

PR #563, opened 2019-10-01
Add plugin website status V1

mikek4233:
This is the V1.  I want to improve the response that you get back and also make the input smarter.


mikek4233:
Sorry, I forgot to add my new library to requirements.txt I will fix and re-request

PR #567, opened 2019-10-02
Added a plugin to convert integers to binary numbers

nataw:
It's very basic. It provides something like the following:

```
~> What can i do for you?
binary 42
101010
~> What can i do for you?
binary
What's your number? 0
0
~> What can i do for you?
binary
What's your number? 1024
10000000000
~> What can i do for you?
binary 65535
1111111111111111
~> What can i do for you?
```


pnhofmann:
> It's very basic.

... but it's nice and really useful! Thanks!

Issue #568, opened 2019-10-03
Plugin for DNS Lookup (Reverse and Forward)


mikek4233:
A plugin that you can put an IP address in and get back the hostname and vice versa

mikek4233:
#572

PR #572, opened 2019-10-03
Reverse and Forward DNS Lookup Plugin

mikek4233:
This plugin does a reverse and forward DNS lookup. (Looks up IP based on hostname and looks up hostname based on IP) #568 

Let me know what you think.

### DIFF
--- a/jarviscli/plugins/dns_lookup.py
+++ b/jarviscli/plugins/dns_lookup.py
@@ -3,49 +3,41 @@ import socket
 from colorama import Fore
 
 
-# a reverse DNS lookup or reverse DNS resolution (rDNS)
-# is the querying technique of the Domain Name System (DNS)
-# to determine the domain name associated with an IP address
-@plugin("dns reverse")
-def hostname_lookup(jarvis, s):
+def ip_lookup(hostname):
+    return str(socket.gethostbyname(hostname))
+
+
+def hostname_lookup(ip):
+    return str(socket.gethostbyaddr(ip)[0])
+
+
+def dns_lookup(jarvis, s, txt, func):
     while True:
-        # asks for an IP address
-        ip = str(jarvis.input("Please input an IP address: "))
+        request = str(jarvis.input("Please input a " + txt + ": "))
         try:
-            # tries to find the hostname for given IP address
-            jarvis.say("The hostname for that IP address is: " +
-                       str(socket.gethostbyaddr(ip)[0]), Fore.BLUE)
-            break
+            if txt == 'ip':
+                jarvis.say("The hostname for that IP address is: " +
+                           func(request), Fore.CYAN)
+                return
+            else:
+                jarvis.say("The IP address for that hostname is: " +
+                           func(request), Fore.CYAN)
+                return
         except Exception as e:
-            # prints error and asks if want to try again
             jarvis.say(str(e), Fore.RED)
-            jarvis.say("Please make sure you are inputing a valid IP address")
+            jarvis.say("Please make sure you are inputing a valid " + txt)
             try_again = jarvis.input("Do you want to try again (y/n): ")
             try_again = try_again.lower()
-            if try_again == 'y':
-                continue
-            else:
-                break
-
-# a "forward" DNS lookup is a lookup of an IP address from a domain name.
+            if try_again != 'y':
+                return
 
 
 @plugin("dns forward")
-def ip_lookup(jarvis, s):
-    while True:
-        hostname = str(jarvis.input("Please input a hostname: "))
-        try:
-            # tries to find the hostname for given hostname
-            jarvis.say("The IP address for that hostname is: " +
-                       str(socket.gethostbyname(hostname)), Fore.BLUE)
-            break
-        except Exception as e:
-            # prints error and asks if want to try again
-            jarvis.say(str(e), Fore.RED)
-            jarvis.say("Please make sure you are inputing a valid IP address")
-            try_again = jarvis.input("Do you want to try again (y/n): ")
-            try_again = try_again.lower()
-            if try_again == 'y':
-                continue
-            else:
-                break
+def ip_lookup1(jarvis, s):
+    dns_lookup(jarvis, s, "hostname", ip_lookup)
+
+
+@plugin("dns reverse")
+def hostname_lookup1(jarvis, s):
+    dns_lookup(jarvis, s, "ip", hostname_lookup)
+
--- a/jarviscli/plugins/dns_lookup.py
+++ b/jarviscli/plugins/dns_lookup.py
@@ -6,7 +6,7 @@ from colorama import Fore
 # a reverse DNS lookup or reverse DNS resolution (rDNS)
 # is the querying technique of the Domain Name System (DNS)
 # to determine the domain name associated with an IP address
-@alias('hostname lookup', 'what is the hostname for')
+@alias('hostname lookup', 'hostname for')
 @plugin("dns reverse")
 def hostname_lookup(jarvis, s):
     while True:
@@ -31,7 +31,7 @@ def hostname_lookup(jarvis, s):
 # a "forward" DNS lookup is a lookup of an IP address from a domain name.
 
 
-@alias('ip lookup', 'what is the ip for')
+@alias('ip lookup', 'ip for')
 @plugin("dns forward")
 def ip_lookup(jarvis, s):
     while True:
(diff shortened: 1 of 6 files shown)


mikek4233:
Hmmm not sure what those errors mean. 

When I ran locally test.sh it came back fine

Any thoughts?

pnhofmann:
> Hmmm not sure what those errors mean.

Most definitely not your bug ;). My current guess: 'what is the ip for' and 'what is the hostname for' basically creates plugin 'what' - then sub-plugin 'is' for 'what' - then sub-plugin 'the' what and so on.
Guess this somehow triggers unknown bug in the PluginManager (code I wrote...).

Plugin names with 5 words is something we - I think  - did not have before. There where times when only 2 words were allowed ;). I will investigate this bug.

However, apart from that: Would prefer just 'ip for' and 'hostname for' for plugin names.

Plugin works - and would have no problem to merge. Thanks!
But if you like, I would make some suggestions how to improve this code - maybe for you to learn. Because in fact 'learning is a big purpose of this project ;).

mikek4233:
Ok I will fix the alias names. 

I would love suggestions - always want to learn more. 

I tried to make the "retry" piece of the code a function as it is duplicated but I couldn't figure out how to either put continue and break in the function without a while loop (I don't think you can) and if I put the while loop in the function then you would need to call a passed-in function which I think you can do but I couldn't get it to work. I'd love some advice!

mikek4233:
I hope its ok I left the names as DNS forward and DNS reverse as that is actually what it is doing

pnhofmann:
No need for removing alias completely! That's no error - and the bug I made, I just fixed (#573).

> I tried to make the "retry" piece of the code a function as it is duplicated but I couldn't figure out how to either put continue and break in the function without a while loop (I don't think you can)

Big ugly - you probably would have to return True/False in your function - and then in the while loop do something like

```
if call_function():
    break
```

But let's rather do this: 

> and if I put the while loop in the function then you would need to call a passed-in function which I think you can do but I couldn't get it to work.

I really love playing with high level functions in Python! So it we define a function like this:

```
def hostname_resolve(ip):
    return str(socket.gethostbyaddr(ip)[0])
```

We can just pass this function to another function. Say:

```
def demo(jarvis, s, some_function):
    print(some_function(s))

demo(hostname_resolve)
```

So whole trick is not to put any () behind hostname_resolve. If you add () you always call a function. If no () - you pass the function around. Which in the context of demo is now called 'some_function' - but can be called just like any other function. 

Does this help?

jarviscli/plugins/dns_lookup.py
```diff
@@ -0,0 +1,51 @@
...
+def hostname_lookup(jarvis, s):
+    while True:
+        # asks for an IP address
+        ip = str(jarvis.input("Please input an IP address: "))
+        try:
+            # tries to find the hostname for given IP address
+            jarvis.say("The hostname for that IP address is: " +
+                       str(socket.gethostbyaddr(ip)[0]), Fore.BLUE)
+            break
+        except Exception as e:
```
pnhofmann: Just noticed: As a rule, never catch 'Exception'.

For example, let's imagine a typo 'Fore.BUE'. Your function will work without showing a error - but whatever you input - valid or invalid IP - it will always say 'Please make sure you are inputing a valid IP'.

Let's look at socket documentation:

https://docs.python.org/3.5/library/socket.html#exceptions

You should make it a:

``except socket.gaierror as e``

Because that's what is returned if ip not found. Additionally you could catch ``OSError`` in a second except clause. This would include e.g. ``socket.timeout`` - and you could tell the user, something really unexpected happened - but IP address might be right.

jarviscli/plugins/dns_lookup.py
```diff
@@ -0,0 +1,51 @@
...
+            # prints error and asks if want to try again
+            jarvis.say(str(e), Fore.RED)
+            jarvis.say("Please make sure you are inputing a valid IP address")
+            try_again = jarvis.input("Do you want to try again (y/n): ")
+            try_again = try_again.lower()
+            if try_again == 'y':
+                continue
```
pnhofmann: Isn't this clause obsolete? Writing:

```
if try_again != 'y':
    break
```

might be more readable.

jarviscli/plugins/dns_lookup.py
```diff
@@ -0,0 +1,51 @@
...
+            else:
+                break
```
pnhofmann: Simply writing:

```
if try_again != 'y':
    break
```

might be more readable.

mikek4233: Definitely

jarviscli/plugins/dns_lookup.py
mikek4233: Makes sense

mikek4233:
@pnhofmann 
I don't totally understand your suggestion, can you please elaborate? 

When I was trying before this is what I tried: 
`
def try_again(func):
    while True:
        try_again = input("Do you want to try again (y/n): ")
        try_again = try_again.lower()
        if try_again == 'y':
            func()
        else:
            break

@plugin("dns reverse")
def hostname_lookup(jarvis, s):
    ip = str(jarvis.input("Please input an IP address: "))
    try:
        jarvis.say("The hostname for that IP address is: " +
                    str(socket.gethostbyaddr(ip)[0]), Fore.BLUE)
    except Exception as e:
        jarvis.say(str(e), Fore.RED)
        try_again(hostname_lookup)`

However, it can't execute the function and just gets stuck in the loop and asks do you want to try again until you put in n which breaks. 

Any thoughts? and thanks for your time and advice!

jarviscli/plugins/dns_lookup.py
mikek4233: According to the documentation, socket.gaierror would only catch errors from address-related errors by getaddrinfo() and getnameinfo(). and i am using neither. 

socket.herror would catch errors from  gethostbyname_ex() and gethostbyaddr() so I think that is the correct one, what do you think?

mikek4233: Ok so after some testing socket.gaierror does catch errors from socket.gethostbyname but neither socket.herror or OSError is catching errors from socket.gethostbyaddr...based on this I will as Exception for now?

pnhofmann:
That's recursion!

Basically

* hostname_lookup calls try_again.
* try_again calls hostname_lookup
* hostname_lookup calls try_again - again
* try_again (2) calls hostname_lookup - again
* and hostname_lookup calls try_again - again again
* (...)

This achieves same effect as a while-loop 1)!

=> Just remove redundant while-loop:
```
def try_again(func):
    try_again = input("Do you want to try again (y/n): ")                                    
    try_again = try_again.lower()                                    
    if try_again == 'y':                                    
        func()
```

 
1) at last up to some point - while ``while True`` runs forever, calling functions repeatedly  within each other is limited to (python) usually ~1000 calls - or you will run out of stack memory...

mikek4233:
I tried to remove the while loop and it still didn't work when I reached the part where try_again calls hostname_lookup it just exits

mikek4233:
I did come up with another way that does work though:
```
def ip_lookup(hostname):
    return str(socket.gethostbyname(hostname))


def hostname_lookup(ip):
    return str(socket.gethostbyaddr(ip)[0])


def dns_lookup(jarvis, s, txt, func):
    while True:
        request = str(jarvis.input("Please input a " + txt + ": "))
        try:
            print(func(request))
            break
        except Exception as e:
            jarvis.say(str(e), Fore.RED)
            jarvis.say("Please make sure you are inputing a valid " + txt)
            try_again = jarvis.input("Do you want to try again (y/n): ")
            try_again = try_again.lower()
            if try_again != 'y':
                break


@plugin("dns forward")
def ip_lookup1(jarvis, s):
    dns_lookup(jarvis, s, "hostname", ip_lookup)


@plugin("dns reverse")
def hostname_lookup1(jarvis, s):
    dns_lookup(jarvis, s, "ip", hostname_lookup1)

```

Any thoughts?

appi147:
You can use return instead of break.

appi147:
Bit unlucky with test:
`
jarviscli/plugins/dns_lookup.py:43:1: W391 blank line at end of file`

mikek4233:
Thanks! Just fixed

pnhofmann:
> Any thoughts?

Really like this code!

mikek4233:
Thank you so much!

pnhofmann:
Strange, don't fully understand why e.g. Github shows football.py as 'added' in this [PR](https://github.com/sukeesh/Jarvis/pull/572/files#diff-7be466760c81f83fef8fed07391f986f). But doesn't really matter ;).

Thanks for creating this plugin!

mikek4233:
I was a bit confused by that as well. 

Glad to help! Now on to the next one!

PR #574, opened 2019-10-05
bmi plugin: Fix ZeroDivisionError for metric system

alexandr7035:
Fix ZeroDivisionError if 0m height was entered


pnhofmann:
Thanks!

PR #580, opened 2019-10-09
Update helper.py

Aman-am:
Minor fix


appi147:
What do you think of such changes @pnhofmann

appi147:
I know its October(Hacktoberfest), but it will be better if you create a new PR with actual content

PR #585, opened 2019-10-11
Remove psutil for unix systems

pnhofmann:
psutil requires gcc + python devel packages on linux/macos

* Install psutil only on windows
* Revert to old 'free ram' command for non-windows systems
* Fixes docker build

### DIFF
--- a/installer/requirements.txt
+++ b/installer/requirements.txt
@@ -11,7 +11,7 @@ mock
 nltk
 pick
 pluginmanager
-psutil
+psutil; sys_platform == 'win32'
 pycricbuzz
 pydoc-markdown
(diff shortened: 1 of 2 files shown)


appi147:
oops

pnhofmann:
I should have noticed when reviewing your "add windows support"-PR...

appi147:
Actually it worked on my Linux machine, so....

pnhofmann:
You probably have python-dev package installed - I do, too - so worked for me, too. Actually bit of a pity, because your 'free ram' plugin is just a lot prettier than 'free'-output - but we can't guaranty everyone has python-dev installed. So if you are ok with this, I would say we just let it be ``os.system('free -lm')`` for linux/macos for now.

PR #587, opened 2019-10-12
Project euler plugin

alexandr7035:
Plugin that gives access to project euler problems via jarvis (yet another useless plugin :)))
Info: https://www.projecteuler.net/about

### DIFF
--- a/jarviscli/plugins/project_euler.py
+++ b/jarviscli/plugins/project_euler.py
@@ -32,7 +32,9 @@ class Euler():
         self.jarvis.say('1) Get the problem by number')
         self.jarvis.say('2) Get random problem')
         self.jarvis.say('3) Info')
-        choice = self.jarvis.input_number('Your choice: ', rtype=int, rmin=1, rmax=3)
+        # Just do nothing
+        self.jarvis.say('4) Exit')
+        choice = self.jarvis.input_number('Your choice: ', rtype=int, rmin=1, rmax=4)
 
         if choice == 1:
             problem_number = self.jarvis.input_number('Please, enter the desired number: ', rtype=int)
--- a/jarviscli/plugins/project_euler.py
+++ b/jarviscli/plugins/project_euler.py
@@ -37,11 +37,14 @@ class Euler():
         choice = self.jarvis.input_number('Your choice: ', rtype=int, rmin=1, rmax=4)
 
         if choice == 1:
-            problem_number = self.jarvis.input_number('Please, enter the desired number: ', rtype=int)
+            problem_number = self.jarvis.input_number('Please, enter the desired number: ', 
+                                                       rtype=int,
+                                                       rmin=1,
+                                                       rmax=self.last_problem_id)
             self.get_problem_by_number(problem_number)
         elif choice == 2:
             # Generate random number
-            problem_number = random.randint(1, self.last_problem)
+            problem_number = random.randint(1, self.last_problem_id)
             self.get_problem_by_number(problem_number)
         elif choice == 3:
             self.show_info()
--- a/jarviscli/plugins/project_euler.py
+++ b/jarviscli/plugins/project_euler.py
@@ -91,7 +91,7 @@ class Euler():
         if dirty_text[0] != '':
             dirty_text.insert(0, '')
         if dirty_text[len(dirty_text) - 1] != '':
-            dirty_text.insert(len(dirty_text) - 1, '')
+            dirty_text.insert(len(dirty_text), '')
 
         # Form string again
         sep = '\n'
--- a/jarviscli/plugins/project_euler.py
+++ b/jarviscli/plugins/project_euler.py
@@ -5,6 +5,7 @@ import random
 from bs4 import BeautifulSoup
 
 
+@alias('project euler')
 @require(network=True)
 @plugin('euler')
 class Euler():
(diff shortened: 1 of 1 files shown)


alexandr7035:
The only issue i have now: i can't get rid of this warning on start :(
```
/home/alexandr7035/.prog.d/Jarvis/env/lib/python3.7/site-packages/pluginmanager/plugin_manager.py:164: UserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("lxml"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.

The code that caused this warning is on line 164 of the file /home/alexandr7035/.prog.d/Jarvis/env/lib/python3.7/site-packages/pluginmanager/plugin_manager.py. To get rid of this warning, pass the additional argument 'features="lxml"' to the BeautifulSoup constructor.

  self.plugins.append(klass())
`


`

**Comment by alexandr7035** [comment]
Even addig `features="lxml"` to bs4 constructor

**Comment by pnhofmann** [comment]
Nice plugin!

> The only issue i have now: i can't get rid of this warning on start :(

Hmm, side effect of the plugin manager...

You import BeautifulSoup, PluginManager sees class 'BeautifulSoup' and creates one instance to check if BeautifulSoup is a plugin (which it isn't).
 This happens long before you can even set ``features="lxml"``.

Workaround:

```
diff --git a/jarviscli/plugins/project_euler.py b/jarviscli/plugins/project_euler.py
index 59b6eb2..c9d3c53 100644
--- a/jarviscli/plugins/project_euler.py
+++ b/jarviscli/plugins/project_euler.py
@@ -2,7 +2,7 @@ from plugin import plugin, alias, require
 from colorama import Fore, Style
 import requests
 import random
-from bs4 import BeautifulSoup
+import bs4
 
 
 @alias('project euler')
@@ -60,7 +60,7 @@ class Euler():
             return
 
         # Use bs4 to parse the page
-        soup = BeautifulSoup(page.content, 'html.parser')
+        soup = bs4.BeautifulSoup(page.content, 'html.parser')
 
         # We are only interested in the block with id 'content'
         content = soup.find('div', id='content')
@@ -123,7 +123,7 @@ class Euler():
         # Use bs4 to parse the page with recent problems
         url = self.project_url + '/recent'
         page = requests.get(url)
-        soup = BeautifulSoup(page.content, 'html.parser')
+        soup = bs4.BeautifulSoup(page.content, 'html.parser')
 
         # We need only the table with recent problems
         problem_table = soup.find('table', id='problems_table')
```

Basically just import bs4 - then pluginmanager won't see class 'BeautifulSoup', won't create an instance - and no message is printed!
Which of course you can't really know - I should probably document this issue somewhere...

But what you definitely should fix is PEP8 so check does not fail ;).

```
$ flake8 --select E,W --max-line-length=140 --ignore E722,W503,W504,E128 jarviscli/

jarviscli/plugins/project_euler.py:84:34: E226 missing whitespace around arithmetic operator

jarviscli/plugins/project_euler.py:85:69: E226 missing whitespace around arithmetic operator

jarviscli/plugins/project_euler.py:93:38: E226 missing whitespace around arithmetic operator

jarviscli/plugins/project_euler.py:94:46: E226 missing whitespace around arithmetic operator

jarviscli/plugins/project_euler.py:111:141: E501 line too long (172 > 140 characters)

jarviscli/plugins/project_euler.py:115:141: E501 line too long (175 > 140 characters)

The command "flake8 --select E,W --max-line-length=140 --ignore E722,W503,W504,E128 jarviscli/" exited with 1.
```

**Comment by alexandr7035** [comment]
Great explanation, thanks!

**Comment by alexandr7035** [comment]
Oh, sorry for pep8. I used another utility (pycodestyle) that showed only warnings that i have some lines longer than 79 characters (so I decided to ignore it). I'll fix it :)

**Comment by alexandr7035** [comment]
A question that applies not only to this PR... Should I continue to use prefixes like "euler plugin: ..." in my commits?(for clear history) Or there is no such need?

File: jarviscli/plugins/project_euler.py [hunk 1]
```diff
@@ -0,0 +1,140 @@
...
+
+        # Get problem text
+        problem_div = content.find('div', class_='problem_content')
+        dirty_text = problem_div.get_text()
+
+        # FIXME !!!
+        # Not obvious way to remove duplicate newlines
+        # But I still don't get along with regular expressions :)
+        # FIXME !!!
+        dirty_text = dirty_text.split('\n')
```
pnhofmann:
```suggestion
        dirty_text = dirty_text.splitlines()
```

Better to use splitlines, because Windows it's '\r\n'  - splitlines works in all cases.

jarviscli/plugins/project_euler.py
```diff
@@ -0,0 +1,140 @@
...
+        i = 0
+        while i < len(dirty_text) - 1:
+            if dirty_text[i] == '' and dirty_text[i] == dirty_text[i + 1]:
+                del dirty_text[i]
+            else:
+                i += 1
```
pnhofmann: Hmm... Should be something like

```
import re
re.sub('\n\n\n*', '\n\n', dirty_text)
```
(before split lines) - so basically replace any 2 newlines followed by any number of newlines (the star) with just 2 newlines,
Which again has the \r\n issue. So maybe:

```
re.sub('\r?\n\r?\n(\r?\n)*', '\n\n', dirty_text)
```

Which adds the optional (the question-mark) \r.

But I am OK with your code, too!

alexandr7035:
Replacing split() with splitlines() found and fixed another small bug :)

alexandr7035:
> If you e.g. look at Linux kernel history (https://github.com/torvalds/linux/commits/master?after=da94001239cceb93c132a31928d6ddc4214862d5+34) all commits look this way.

817k commits, .............................................

pnhofmann:
Let's merge ;) Thanks!

alexandr7035:
> Let's merge ;) Thanks!

thank you too!

PR #589, opened 2019-10-12
Added speak argument to jarvis.say

### LINKED ISSUE #588: Jarvis.say should not always speak out
The current implementation of `jarvis.say` speaks out the text whenever speech option is enabled. Instead, long or specially formatted text (say long lists, tables or ASCII art) should not be spoken out even if speech is enabled. 

To accommodate this, I am adding an optional bool argument to `jarvis.say`, that controls whether text should be spoken out loud or not. This is by default `True`, to maintain the current functioning of plugins.

NilayMajorwar:
Added optional bool `speak` argument to control whether text should be spoken or not (to avoid jarvis speaking out long texts, tables or things like ASCII art).
Closes #588

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -15,7 +15,7 @@ in the utilities-package should be implemented here.
 
 ## say
 ```python
-JarvisAPI.say(self, text, color='')
+JarvisAPI.say(self, text, color='', speak=True)
 ```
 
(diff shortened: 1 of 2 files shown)


pnhofmann:
Thanks!

PR #590, opened 2019-10-12
Initial battery plugin implementation without upower

brennop:
Adds a fallback for Linux systems without upower using [this](https://askubuntu.com/a/309146).

### DIFF
--- a/jarviscli/PluginManager.py
+++ b/jarviscli/PluginManager.py
@@ -259,7 +259,13 @@ class PluginDependency(object):
     def _check_native(self, values, plugin):
         missing = ""
         for native in values:
-            if not executable_exists(native):
+            if native.startswith('!'):
+                # native should not exist
+                requirement_ok = not executable_exists(native[1:])
+            else:
+                requirement_ok = executable_exists(native)
+
+            if not requirement_ok:
                 missing += native
                 missing += " "
 
--- a/jarviscli/plugins/battery.py
+++ b/jarviscli/plugins/battery.py
@@ -1,4 +1,5 @@
 import subprocess
+from utilities.GeneralUtilities import executable_exists
 from plugin import plugin, require, LINUX, WINDOWS
 
 VALID_OPTIONS = ['status', 'vendor', 'energy', 'technology', 'remaining']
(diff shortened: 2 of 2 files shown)


pnhofmann:
Hi!
good idea - works on my laptop!

Nevertheless, you should check if this file ``/sys/class/power_supply/BAT0/status'`` actually exist before trying to read. Or catch errors while reading ;).

E.g. if someone tries to call 'battery' from a Desktop-PC without battery. Or if for some other reason this path does not exist 1), code should not throw an error like this:

```
Jarvis' sound is by default disabled.
In order to let Jarvis talk out loud type: enable sound
Type 'help' for a list of available actions.
106 plugins loaded 6 plugins disabled. More information: status

~> Hi, what can I do for you?
battery
cat: /sys/class/power_supply/BAT0/status: No such file or directory
Some error occurred, please open an issue on github!
Here is error:

  File "/home/phil/git/Jarvis/jarviscli/CmdInterpreter.py", line 188, in try_do
    do(self, s)
  File "/home/phil/git/Jarvis/jarviscli/plugin.py", line 209, in run
  File "/home/phil/git/Jarvis/jarviscli/plugins/battery.py", line 50, in battery_LINUX
    battery_text.append("Status: " + get_battery_info("status"))
  File "/home/phil/git/Jarvis/jarviscli/plugins/battery.py", line 47, in get_battery_info
    return subprocess.check_output(["cat", battery_dir + info]).decode("utf-8")[:-1]
  File "/usr/lib64/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib64/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['cat', '/sys/class/power_supply/BAT0/status']' returned non-zero exit status 1.
```


1) Note: From your askubuntu link:
> UPDATE: Linux 3.19 and onwards, we should look at the following directory -> /sys/class/power_supply/BAT1/

Which isn't the case for me - openSUSE 15.1 on 5.2.8-1-vanilla - but guess it is possible, that it is called 'BAT1' on some machines


But really like your commit! If you could add some ``if path exists``, your PR would be perfectly ;).

brennop:
You're right. When I have some time I'll look into adding a check for the file.

> (*) Note: From your askubuntu link:
>
>    UPDATE: Linux 3.19 and onwards, we should look at the following directory -> /sys/class/power_supply/BAT1/
>
> Which apparently also isn't the case for me - openSUSE 15.1 on 5.2.8-1-vanilla - but guess it is possible, that it is called 'BAT1' on some machines

It would be nice to have the script find out if it's called BAT0, BAT1, BATwhatever. I'll try to add that.

pnhofmann:
Added BAT0, BAT1, ... search and merged - Thanks for creating this feature!

PR #594, opened 2019-10-18
Update .travis.yml

appi147:


appi147:
On windows default python was 3.8, so i changed it back to 3.7.4 and since env was for python37 it wasn't able to find path.

Issue #537, opened 2019-07-30
Search google for questions Jarvis doesn't have a plugin for


CremBluRay:
I noticed that Jarvis doesn't understand google-search type questions, so I was wondering if I could work on a plugin that searched google and parsed the results.

CremBluRay:
I was thinking I could use something like serpwow to do it. It would be really simple and might improve natural language understanding of the AI.

appi147:
Go ahead!

CremBluRay:
thanks! I'm on it!

PappaStalin:
Because this feature has gone a wile without being implemented, and it seems like it would add a lot of extensibility to Jarvis, I'm going to see If I can get something like this to work.

PappaStalin:
I think that this is a great idea, and that it would add a lot of extensibility to Jarvis. However, this issue has not been resolved in 3 months. Do you mind if I try to add this feature?

PappaStalin:
I've made a quick and dirty plugin that you can find here: https://pastebin.com/QvmFQWym

pnhofmann:
> Do you mind if I try to add this feature?

Sure - go ahead.

> I've made a quick and dirty plugin that you can find here: https://pastebin.com/QvmFQWym

Yeah, something like this...

https://pypi.org/project/google/

> I was thinking that, instead of returning "I could not identify your command" when Jarvis runs into an error, it prompts the user if they want to google the question. this could also be done explicitly with the search command.

As an optional feature, which needs to be enabled manually, I'd be OK with that.

But don't like the idea of stuff getting sent to Google by default.

PappaStalin:
I've got the google plugin up and running, the only issue is the BS4 raises a UserWarning (No parser was explicitly specified) when Jarvis is run. I know that this is because of the pluginmanager module, but I don't really know why this is the case or how to hide the error. 

the code for my repo can be find here: https://github.com/PappaStalin/Jarvis

pnhofmann:
Hmm, yes - known issue I should really document somewhere (and this time I hopefully won't forget again, Sorry ;) ).

Only import
```
import BeautifulSoup
```

and use as ```BeautifulSoup.bs4```. Otherwise PluginManager crease bs4-instance to check if bs4 is a plugin - and that prints UserWarning.

pnhofmann:
Nice @PappaStalin  - another issue to close ;)

#537

PR #605, opened 2019-10-27
Google Search Plugin for Jarvis

PappaStalin:
A basic command line extension to Jarvis that finds Google's default answer to any question. It can be used by simply typing google and then the thing you want to google for. 

resolution for issue #537


pnhofmann:
Thanks!

Issue #606, opened 2019-10-29
New Feature- Open Browsers through Jarvis for more precise searching


akshgpt7:
There are a lot of things that you won't be able to search through the command line directly. So, a command to open browser windows directly through Jarvis would make it more interesting altogether.

pnhofmann:
Hi!

* There's open website (Just noticed: Currently broken ;( => I'll push a fix later.)
* There is google (Soon ;) => #605)

If I correctly understand, what you want is something like an extension for ``google``:

* Google in webbrowser

Which instead of printing short summary opens webbrowser on google? This would be a nice plugin. You would like to implement it?

akshgpt7:
Yes, you got it right. I'll work on the command ```open browser``` .

pnhofmann:
#616

PR #610, opened 2019-11-02
added browser plugin

akshgpt7:
Plugin mentioned in issue #606 .


pnhofmann:
Hi!
solid implementation for opening google.com ;).
Just a suggestion: Maybe also allow queries like:
```browser Jarvis```
Which would open something like:
https://www.google.com/search?q=jarvis

akshgpt7:
Doesn't the ```google``` command do query searches already?

pnhofmann:
Yes, but ``google`` only returns prints one sentence (summary) on CLI - and does not open web browser to show all search results.

akshgpt7:
Sure, I'll add this feature :)

PR #612, opened 2019-11-03
Fixed open website plugin

akshgpt7:
Hey! The ```open website``` plugin works just fine now.
Plus, I added some more popularly visited websites to the list.

I'll edit the browser plugin separately and add the keyword searching feature.


pnhofmann:
Thanks!

Issue #499, opened 2019-05-25
GUI calculator


CDR-Isaac:
I've been thinking towards a more graphical interface of Jarvis, just like in the Iron Man movies, and I know we could use a GUI calculator. Here is the code: 

   from tkinter import *
   import re

   class Calculator:
       def __init__(self, master):
           self.master = master
           master.title("Calculator")

           self.screen = Text(master, state='disabled', width=30, height=3, background="yellow", foreground="blue")

           self.screen.grid(row=0,column=0,columnspan=4,padx=5,pady=5)
           self.screen.configure(state='normal')
   
           self.equation = ''

           b1 =  self.createButton(7)
           b2 = self.createButton(8)
           b3 = self.createButton(9)
           b4 = self.createButton(u"\u232B",None)
           b5 = self.createButton(4)
           b6 = self.createButton(5)
           b7 = self.createButton(6)
           b8 = self.createButton(u"\u00F7")
           b9 = self.createButton(1)
           b10 = self.createButton(2)
           b11 = self.createButton(3)
           b12 = self.createButton('*')
           b13 = self.createButton('.')
           b14 = self.createButton(0)
           b15 = self.createButton('+')
           b16 = self.createButton('-')
           b17 = self.createButton('=',None,34)
          
           buttons = [b1,b2,b3,b4,b5,b6,b7,b8,b9,b10,b11,b12,b13,b14,b15,b16,b17]   

           count = 0

           for row in range(1,5):
               for column in range(4):
                   buttons[count].grid(row=row,column=column)
                   count += 1

           buttons[16].grid(row=5,column=0,columnspan=4)

       def createButton(self,val,write=True,width=7):

           return Button(self.master, text=val,command = lambda: self.click(val,write), width=width)

       def click(self,text,write):
           if write == None:

               if text == '=' and self.equation:
                   self.equation= re.sub(u"\u00F7", '/', self.equation)
                   print(self.equation)
                   answer = str(eval(self.equation))
                   self.clear_screen()
                   self.insert_screen(answer,newline=True)
               elif text == u"\u232B":
                   self.clear_screen()

           else:
               self.insert_screen(text)

       def clear_screen(self):
           self.equation = ''
           self.screen.configure(state='normal')
           self.screen.delete('1.0',END)

       def insert_screen(self, value,newline=False):
           self.screen.configure(state='normal')
           self.screen.insert(END,value)
           self.equation += str(value)
           self.screen.configure(state ='disabled')

   root = Tk()
[...]

PR #621, opened 2019-11-10
GUI calculator added

akshgpt7:
Added a GUI calculator to Jarvis.
The calculator can be opened by the command ```calculator``` .

Issue #499


jarviscli/plugins/calculator.py
```diff
@@ -0,0 +1,292 @@
+import tkinter as tk
```
pnhofmann: Hmm, tkinter may not be installed. We should add some kind of check if this import fails (I'll think of something).

Have some graphical glitches - probably because my i3 desktop is messing with windows size... I'll check with a "regular" window manger later ;).

First look code looks good to me.

akshgpt7:
- I've fixed the window size messing on i3.

- But Tkinter is a standard python module. Anyway, Do you want me to add a check for the import inside the plugin code itself, and install the module if not present?
 Or I can add it to the list of modules (requiremets.txt) to be installed while setting up Jarvis.

akshgpt7:
- I've fixed the window size messing on i3.

- But Tkinter is a standard python module and is installed along with python. Anyway, Do you want me to add a check for the import inside the plugin code itself, and install the module if not present?
 Or I can add it to the list of modules (requiremets.txt) to be installed while setting up Jarvis.

pnhofmann:
Most python installations include tkinter - but you can't be 100% sure. Linux-Distributions tkinter usually have a unique package for tkinter - e.g. Ubuntu 'python-tk'.
afaik tkinter can't be installed wih pip (requirements.txt), bceause tkinter must be build within the python installation.

So I have just added small check to avoid problems.

pnhofmann:
Thanks!

PR #632, opened 2019-11-27
Minor fixes and improvements

pnhofmann:


appi147:
Hey are we going py38 only?

pnhofmann:
@appi147 
Sorry, that was unclear of me. No, wouldn't even see a benefit from py38 only ;).

Just noticed, two issues with python3.8: Comparing Strings with 'is' throws syntax errors (and in fact comparing strings with is was wrong even before) and deprecated method 'platform.dist()' was removed completely and needed replacement.

That was the first commit.

Afterwards noticed some view other minor issues, I wanted to fix and just pushed; would probably have been better to send them in new PR...

appi147:
Just messing with you!!
Py38 has cool features tho

PR #640, opened 2019-12-14
working directory should be returned back to original state after test

hawkeye359:
added cd .. at the end of the file test.sh to send back to original working directory after the test is complete
now due to the `cd jarviscli/` command the cursor stays at the folder jarviscli folder even if the test is complete

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -11,3 +11,4 @@ echo "checking for unit test"
 python3 -m unittest discover
 echo "unit tests checked"
 echo ""
+cd ..


pnhofmann:
Are you running ./test.sh with ``source`` or ``. ``? Because calling cd inside a script otherwise does not effect working directory outside the script.

Sure, I have no problem resetting work directory at end of the script - asking more out of curiosity ;).

hawkeye359:
I am very thankful to you because this was my first ever pull request on any open source project.

I was running it with `source` and it was leaving me at `jarviscli` and I tested it again after your comment without my change `cd ..`. But the result was same.
But after the change it is working correctly.

Additionally thanks for merging my commit. It means me a lot :).

pnhofmann:
> Additionally thanks for merging my commit. It means me a lot :).

Sure - thank you for your contribution ;).

PR #641, opened 2019-12-14
add plugins

Baocaifeng:


Workflow.md
```diff
@@ -0,0 +1,10 @@
+# Workflow
```
pnhofmann: What is this file for?

custom/speech.py
```diff
@@ -0,0 +1,248 @@
+from plugin import plugin, require, LINUX
```
pnhofmann: For release, please move to jarviscli/plugins folder

custom/speech.py
```diff
@@ -0,0 +1,248 @@
...
+        'speech_installed (install speech_recognition playsound + re-run setup.sh)']
+
+
+#global values
+num = 1
+
+
+@require(native=requirements, platform=LINUX)
+@plugin("open speech")
+def open_speech__LINUX(jarvis, s):
```
pnhofmann: how does this compare to ``jarvicscli/plugins/voice_control.py``?

(And yes: voice_control.py is still in a really early stage, I have multiple ideas to improve but didn't spent much time implementing yet ;) )

custom/speech.py
```diff
@@ -0,0 +1,248 @@
...
+            speak = "I have been created by Human."
+            assistant_speaks(speak) 
+            return
+  
+        elif "geeksforgeeks" in input:# just 
+            speak = """Geeks for Geeks is the Best Online Coding Platform for learning."""
+            assistant_speaks(speak) 
+            return
+  
+        elif "calculate" in input.lower():
```
pnhofmann: Doesn't this re-implement existing functionality (e.g. open, search web, ...) respectively should be a standalone plugin (e.g. wolfram search)?

jarviscli/plugins/img2pdf.py
```diff
@@ -0,0 +1,31 @@
+from plugin import plugin
```
pnhofmann: img 2 pdf / pdf 2 image looks good to me

custom/speech.py
Baocaifeng: hello, this is the first time that I try to contribute to an open project with many people, sorry about my mistakes ! I'm doing this plugin to satisfy the requirements of my teacher, and, I tried "hear"  plugin, it did work with me (I guess the problem is caused by the driver of SoundCard), I can't solve with this problem and I don't really understand voice_control.py.  I made this plugin "speech", it is an independent plugin like the other plugins, it can recognize human voice quite well. I don't know what I'm saying, maybe you have already found some unperfects about voice control, I'm finishing speech in quite tight time without understanding all the code of Jarvis, emmmh, I don't know what is going to do........maybe you can give me some advice

custom/speech.py
Baocaifeng: yeah, emmmm, emmmmm, emmmm, emmm, how can I say, **speech** is a plugin that can act as **jarvis teminal**, that is to say, it can execute **standard jarvis command** (like battery status), so this means, **speech** plugin can call other plugins like jarvis command **google**  or **calculate**. But the problem is that I what to get the **output** of the called plugins and say it out. If I use **jarvisAPI.eval(command)**, I can't get the output in memory, beacause all the outputs are in jarvis termimal (using print_say or jarvis.say).  My flow is: jarvis calls pluginA ---> pluginA calls pluginB --> pluginB return results to pluginA --> pluginA does something else with getting results --> pluginA can closed jarvis terminal and exit all the processing. 
So it's a little difficult for me to use google/search of jarvis, because I don't understand all the source code and jarvis are designed as independent plugins. to make things easily, I created sub-fuction search/open which I just use in speech.

emmmm, I'm currently a student and not very experienced.....maybe you can give me some guidance ^^

pnhofmann:
> hello, this is the first time that I try to contribute to an open project with many people, sorry about my mistakes ! 

No need to apologize - we are all learning ;).

> If I use jarvisAPI.eval(command), I can't get the output in memory, beacause all the outputs are in jarvis termimal 

Do you really need output of commands? Sure, I could add such a functionality, but 'text to speech' is already implemented in Jarvis (``jarvis.enable_voice()``). 

I'll try to take a closer look at this whole voice control stuff tomorrow ;).

Baocaifeng:
Hello, Thank you!
Yes, it's better to add a function to get the **output**. Yes, I can use enable_voice to say the **output**, but the problem is that I'm studying in France..........**google** plugin can **get the content** in French (according to my location) but **say it in English**.................

Issue #635, opened 2019-12-05
[NEW FEATURE] web page to pdf


baptoutiego:
Hello, 
I would like to add a feature on linux to convert html web page to pdf with the pdfKit library.

pnhofmann:
#642

PR #642, opened 2019-12-14
new htmltopdf plugin

baptoutiego:
this is the implementation of the plugin htmltopdf, link to issue [635](https://github.com/sukeesh/Jarvis/issues/635).
I put two subcommands : 

- url , download web page to a pdf file

- file, convert a html file into a pdf


jarviscli/plugins/htmltopdf_url.py
```diff
@@ -0,0 +1,30 @@
...
+class htmltopdf_url:
+    
+    """transform your url page into a pdf file in the Jarvis source directory. type your url as the following:
+    'htmltopdf google.com'
+    The output file will be the following:
+    'google.com.pdf'"""
+
+    def __call__(self, jarvis, s):
+    	
+    	#We have to add the '.' back beacause the Jarvis API removes it
```
pnhofmann: Year, I'll implement a way that Jarvis API e.g. does not remove '.' of something that looks like an URL ;).

pnhofmann:
Hi!
looks good!

Only issue I noticed on first look, please fix Pep8 (https://travis-ci.org/sukeesh/Jarvis/jobs/625015295)

baptoutiego:
Hi,
I finally fixed pep8 with a few more commits ^^

jarviscli/plugins/htmltopdf_url.py
```diff
@@ -0,0 +1,28 @@
...
+@plugin("htmltopdf url")
+class htmltopdf_url:
+    """
+    Transform your url page into a pdf file in the Jarvis source directory. type your url as the following:
+    'htmltopdf google.com'
+    The output file will be the following:
+    'google.com.pdf'
+    """
+    def __call__(self, jarvis, s):
+        # We have to add the '.' back because the Jarvis API removes it
```
pnhofmann: Hi!
created a fix for this issue (#644).

Really like this feature!

Just have two things to say:
* Could we maybe have this three plugins in one file? Think this would be more well-arranged.
* Also, we should probably add 'wkhtmltopdf' as optional dependency (see: https://github.com/sukeesh/Jarvis/blob/master/installer/optional.py#L28). But could also do this myself.

baptoutiego: Hi!
Glad that you like it.
I put these in different files to follow multiwords commands recommendations in the doc. I also wanted to use subcommands documentation for user in the CLI. That is why i created different plugins. But for sur, I can change the code to have one plugin and different conditions on the inputs. And i can do the optional, it shouldn't take too long 😉

pnhofmann: > I put these in different files to follow multiwords commands recommendations in the doc.

That's the way it is supposed to be! Just meant moving this three independent plugins to one file since there is no need for one file / plugin.

baptoutiego: Ok I understood :)  So in the last commits, I put all the plugins in one file, I remove the code which put back the dot since it is solved, and for the optional, i opened issue https://github.com/sukeesh/Jarvis/issues/646. I will make a new PR about it soon ;)

pnhofmann:
Nice! Thanks!

Issue #646, opened 2019-12-18
add 'wkhtmltopdf' as optional dependency 


baptoutiego:
The goal is to  add 'wkhtmltopdf' as optional dependency (see: https://github.com/sukeesh/Jarvis/blob/master/installer/optional.py so we can use htmltopdf plugins.
linked feature : https://github.com/sukeesh/Jarvis/pull/642

pnhofmann:
#647

PR #647, opened 2019-12-19
646 wkhtmltopdf

baptoutiego:
this is the implementtation of the https://github.com/sukeesh/Jarvis/issues/646
add wkhtmltopdf in optional.py


pnhofmann:
Hmm, again this windows / choco bug in travis... Guess this is again server issue of choco, so nothing to worry about.

Anyway; thanks!

PR #648, opened 2019-12-30
New Plugin: Geocoding

aazuspan:
This is a plugin that allows Jarvis to geocode, converting street addresses to latitude and longitude coordinates. This is done using the public geocoding API provided by the US Census Bureau.

### DIFF
--- a/jarviscli/plugins/geocode.py
+++ b/jarviscli/plugins/geocode.py
@@ -33,9 +33,9 @@ class Geocoder:
             Likely to be empty. 
         """
         self.jarvis = jarvis
-        self.input_addr = self.get_input(s)
-        self.cleaned_addr = self.urlify(self.input_addr)
-        req = self.make_request()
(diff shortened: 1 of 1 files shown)


pnhofmann:
Really nice code! I have nothing to add.

Just noticed ToC of census.gov:
```

Attribution

All services, which utilize or access the API, should display the following notice prominently within the application: "This product uses the Census Bureau Data API but is not endorsed or certified by the Census Bureau."
```
https://www.census.gov/data/developers/about/terms-of-service.html

Please add this somewhere (would suggest immediatelly when geocode starts) and I'll merge.

aazuspan:
@pnhofmann Good catch! I've added the disclaimer as requested.

aazuspan:
@pnhofmann Sorry about the late commits. I found a few issues when I started writing tests. Should be okay to merge now, if everything still looks good.

pnhofmann:
All right! Thanks!

PR #650, opened 2020-01-07
Add Geocoding Tests

aazuspan:
This PR adds 4 tests for the geocode plugin. Ensures that inputs are correctly cleaned and that outputs are correctly parsed.


pnhofmann:
Nice ;).

Issue #651, opened 2020-01-10
[FEATURE] Store 'sound enabled' setting


aazuspan:
For people who use always use Jarvis with sound enabled, it would be nice if that setting was remembered after closing Jarvis. I'm planning on implementing this using Jarvis' Memory class. That memory would be checked every time Jarvis is opened and would be updated whenever the user enables or disables sound.

PR #652, opened 2020-01-10
Implement #651: Store 'sound enabled' setting

aazuspan:
Adds feature mentioned in #651. 

### Changes
- Load `enable_voice` setting from memory when JarvisAPI is initialized.
- Set `enable_voice` setting in memory when it is enabled or disabled.
- Add and call method for removing ANSI escape sequences from all `text_to_speech()` inputs
  - If this isn't included, colorama color codes passed to `text_to_speech()` (ie `Jarvis.first_reaction_text`) will be read aloud as m39, m40, etc


pnhofmann:
Thanks!

Issue #653, opened 2020-01-13
[Feature] Voice speed control


aazuspan:
I think it would be great if the user could control the speed of Jarvis's speech when voice is enabled. I'm thinking this could be accomplished by adding some plugins to voice.py so the user could type 'talk faster' or 'talk slower'. Additionally, the user's preferred speed would be saved in Jarvis' memory and would be recalled on startup if voice is enabled. I'll start looking into how exactly to implement this, but I'm open to any suggestions on functionality or implementation.

pnhofmann:
#654

PR #654, opened 2020-01-15
Implement #653: Add speech rate control

aazuspan:
This PR adds speed control plugins mentioned in #653 that allow the user to change Jarvis' speech rate with 'talk faster' and 'talk slower' if voice is enabled. Additionally, speech rate is saved in Jarvis' memory so that it can be recalled between sessions. In order to enable these plugins, some changes were made to the JarvisAPI, to Jarvis and CmdInterpreter, and to the voice utility.  

### Changes
- Add `talk faster` and `talk slower` plugins to `plugins/voice.py`
- Add a `change_speech_rate` method to the JarvisAPI to interface between the plugins and the speech engine
- Recall `rate` when CmdInterpreter is initialized
- Move `enable_voice` recall out of the API and into CmdInterpreter, since it seems more logical to do all settings recall there
- Remove `enable_voice` parameters from Jarvis because it will now be recalled from memory or default to off if not previously set
- Add a `rate` parameter to `create_voice` which is then passed to Voice classes and stored
- Add a setter for `rate` for each Voice class that modifies the value
- Modify the `create` method for each voice class to set the speech rate when the engine is created
- Save `min_rate` and `max_rate` attributes and restrict `rate` inside that range

### DIFF
--- a/jarviscli/CmdInterpreter.py
+++ b/jarviscli/CmdInterpreter.py
@@ -167,7 +167,7 @@ class JarvisAPI(object):
         :param delta: Amount of change to apply to speech rate
         """
         new_rate = self._jarvis.speech.change_rate(delta)
-        self.update_data('speech_rate', new_rate)
+        self.update_data('speech_rate', self._jarvis.speech.rate)
 
     # MEMORY WRAPPER
     def get_data(self, key):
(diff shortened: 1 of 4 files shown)


pnhofmann:
Hi - finally I'm here ;). Yeah, good work! Like this feature, clean git log - I have nothing to add.

Thanks!

PR #661, opened 2020-02-22
Adding a new feature: Bulk Resize

chekoduadarsh:
I have added a new feature which can resize images in bulk with padding or the given square size, It can be highly used when we do deep learning applications to resize and clean the dataset without damaging the dataset

### DIFF
--- a/jarviscli/plugins/bulkresize.py
+++ b/jarviscli/plugins/bulkresize.py
@@ -34,7 +34,8 @@ def bulk_resizer(input_path, output_path, desired_size=32,
         top, bottom = delta_h // 2, delta_h - (delta_h // 2)
         left, right = delta_w // 2, delta_w - (delta_w // 2)
 
-        new_im = cv2.copyMakeBorder(im, top, bottom, left, right,cv2.BORDER_CONSTANT, value=color)
+        new_im = cv2.copyMakeBorder(im, top, bottom, left, right,
+                                    cv2.BORDER_CONSTANT, value=color)
         if rename:
             output_path1 = output_path + "/" + str(img_no) + ".jpg"
             img_no += 1
--- a/jarviscli/plugins/bulkresize.py
+++ b/jarviscli/plugins/bulkresize.py
@@ -12,7 +12,7 @@ def bulk_resizer(input_path, output_path, desired_size=32,
     filename = os.listdir(input_path)
     filepath = []
     for name in filename:
-        path = input_path+"/"+name
+        path = input_path + "/" + name
         if os.path.isfile(path):
             filepath.append(path)
     for im_pth in filepath:
--- a/jarviscli/plugins/bulkresize.py
+++ b/jarviscli/plugins/bulkresize.py
@@ -9,15 +9,21 @@ def bulk_resizer(input_path, output_path, desired_size=32,
                  color=[0, 0, 0], rename=True):
     img_no = 0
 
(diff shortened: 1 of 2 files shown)


jarviscli/plugins/bulkresize.py
```diff
@@ -0,0 +1,70 @@
...
+import glob
+
+from plugin import plugin
+
+
+def bulk_resizer(input_path, output_path, desired_size=32,
+                 color=[0, 0, 0], rename=True):
+    img_no = 0
+
+    onlyfiles = os.listdir(input_path)
```
pnhofmann: os.listdir definitely includes directories. Maybe filter with ``os.path.isfile``?

jarviscli/plugins/bulkresize.py
```diff
@@ -0,0 +1,70 @@
...
+    # print(onlyfiles)
+    for im_pth in onlyfiles:
+        try:
+            im = cv2.imread(im_pth)
```
pnhofmann: im_pth does at this point not contain the full path - but only the filename (e.g. jarvis.png).
=> This line will fail for all images not in working directory, you should prepend input_path.

Apropos, there is even in python stdlib: 
https://docs.python.org/3/library/os.path.html#os.path.join

jarviscli/plugins/bulkresize.py
```diff
@@ -0,0 +1,70 @@
...
+               ' data collection process')
+    jarvis.say('Enter the path of directory with images to be resized :')
+    path1 = jarvis.input()
+    jarvis.say('Should I rename them to non repeating whole number series?',
+               ' y/n :')
+    rename = jarvis.input()
+    jarvis.say('Enter the path of output directory :')
+    path2 = jarvis.input()
+    jarvis.say("Enter the target size :")
+    size = jarvis.input_number()
```
pnhofmann: input_number default returns float - which is probably not what you want. Add parameter ``rtype=int``.

jarviscli/plugins/bulkresize.py
```diff
@@ -0,0 +1,70 @@
...
+               'size with padding !!!! Specially designed for Deep Leanring',
+               ' data collection process')
+    jarvis.say('Enter the path of directory with images to be resized :')
+    path1 = jarvis.input()
+    jarvis.say('Should I rename them to non repeating whole number series?',
+               ' y/n :')
+    rename = jarvis.input()
+    jarvis.say('Enter the path of output directory :')
+    path2 = jarvis.input()
+    jarvis.say("Enter the target size :")
```
pnhofmann: If I read this correctly, does (currently) support only square images?

chekoduadarsh: Yes, this library is currently supporting square images!! I was mainly doing this for my deep learning project so I just made small adaptations to make it a feature in Jarvis

jarviscli/plugins/bulkresize.py
chekoduadarsh: Done

jarviscli/plugins/bulkresize.py
chekoduadarsh: Added

jarviscli/plugins/bulkresize.py
chekoduadarsh: Done

pnhofmann:
Thanks! So let's merge ;).

PR #667, opened 2020-03-07
Fix scenario when cd fails before other commands

northernSage:
Change 'python -m unittest discover' call to run in a Sub-shell.

if sequences such as 'cd dir; do something; cd ..', happen to fail (e.g. if dir was deleted or user lacks permission to access it), 'do something' will be executed in the wrong directory producing unwanted results.

These changes prevent this from happening by using a subshell.


pnhofmann:
Thanks

PR #671, opened 2020-03-27
New Feature: History

CSenshi:
As you can see from plugin name, it's feature is to give you some historical fact.
You can use current plugin as follows: 
`history <keyword> <event> <day> <month>`

- keyword - Program uses special keywords to easily identify your query. For now it has only one keyword 'today' that results in getting fact that happened on this day

- event   - Argument used to specify historical fact type, which can be one of the following: 'births', 'deaths', 'events'

- month   - Specify month

- day     - Specify day

if month,event or day is not passed as argument it will be randomized
Example

1. 'history 25 march birth' - birth of a random person on 25th of March
2. 'history today' - random type of event that happened on the present day
3. 'history 23' - random event that happened on random month but day is 23

### DIFF
--- a/jarviscli/plugins/history.py
+++ b/jarviscli/plugins/history.py
@@ -1,4 +1,5 @@
 from plugin import plugin, require
+import random
 
 @require(network=True)
 @plugin('history')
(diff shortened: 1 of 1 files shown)


pnhofmann:
Nice! Will take a closer look tomorrow, but looks really good!

Probably some attribution would be nice - something like "Data taken from wikipedia (CC BY-SA 3.0) with use of http://history.muffinlabs.com" in docstring; but probably not even necessary.

Yes, really like this feature, really readable code.

CSenshi:
> Nice! Will take a closer look tomorrow, but looks really good!
> 
> Probably some attribution would be nice - something like "Data taken from wikipedia (CC BY-SA 3.0) with use of http://history.muffinlabs.com" in docstring; but probably not even necessary.
> 
> Yes, really like this feature, really readable code.

Done! ✌️

pnhofmann:
Thanks!
Sorry, took a week (sort of forgot ;) ).

PR #677, opened 2020-04-14
New Feature: coronavirus status

noam-stein:
This feature will add the  correct coronavirus status:
example:
image

If the user does supply a country the result will be worldwide:
image

If the user enters a country that doesn't appear on the API the result we be as followed:
image


pnhofmann:
Yeah - something up to date. Let's hope we will be able to remove this plugin at some point when corona is controlled and this plugin just not necessary any more. But for now and to check when this all will be over - definitely useful - thanks.

PR #680, opened 2020-04-25
Updated keywords list

shoniavika:
New user might type "check forecast in {location} for a week". So adding 'a' and 'week' to keywords list will make forecast call more user-friendly.


pnhofmann:
Well, at some point we'll probably need to re-think completely how "keywords" are handled. But for now will merge; thanks.

Issue #674, opened 2020-04-07
New Feature: plugin to create new_plugin.py with template


KonstantinosLolos:
Hello! I have been toying around with Jarvis on my computer and I would really like to contribute.
I was messing around with some ideas and I thought about adding a plugin that creates a new_plugin.py file with the Hello World  plugin sample you provide and some extra documentation about the different options a developer has.
You guys have created really helpful info on how to contribute and create new plugins and since this is a plugin based System I thought it would be nice to apply the info you provide to a template! Any thoughts?

pnhofmann:
Sounds good to me!

KonstantinosLolos:
Hello again! I would like your opinion on a matter. I have been able to test my code in both MAC os and LINUX. The only difference between them and the reason i use the require tag is that in linux i have not been able to open the plugin after it has been created (the open command uses terminal based editors and not the default editor of the user which is not practical).

The question is: Should I remove the 'open' feature so the plugin is not that big or are you okay with the current situation?
Thanks!

KonstantinosLolos:
Looks like thats what it does! Thanks for the tip! I added that in and made some more practical changes. I shall proceed with a pull request for review.

KonstantinosLolos:
Not adding anything else for the time being. Should we consider issue closed for now? :)

pnhofmann:
Yeah, forgot closing this one.

Hint: Nice github-feature, if you add issue-id '#674' to one of your commit-message github auto-closes related issues while merging.

PR #684, opened 2020-04-30
Add create_plugin.py plugin 

KonstantinosLolos:
This plugin is designed to make it easier for a developer to start programming and experimenting with Jarvis plugins. It is functional for MAC and LINUX operating systems.
The format of the template which is created is strictly based on the instructions given in the url attached in README.md.

PS: if you decide to merge, might i suggest to mention the plugin in the README file under plugins compartment.
#674


pnhofmann:
Thanks!

Issue #663, opened 2020-02-24
Changing the voice of assistant


chekoduadarsh:
Hey,
 When I was checking up the voice mode of Jarvis, it wasn't pleasant to listen it was super robotic and non-human like. Proposing a migration from `pyttsx` to a combination of `gtts` and `playsound` which gives a way better humanitarian approach!!

pnhofmann:
Sure!
Just, would prefer to let the pyttsx code in Jarvis and add gtts as an alternative backend (and make the user choice if he wants to send stuff to google or not).

``jarviscli/utilities/voice.py`` does implement 3 backend anyway - one for Windows, one for Mac and one for Linux. On Mac pyttsx isn't even used ;).

chekoduadarsh:
> Just, would prefer to let the pyttsx code in Jarvis and add gtts as an alternative backend (and make the user choice if he wants to send stuff to google or not).
Will do it

OneHighCloud:
I am working on Win10 with the code. In voice.py, I made some changes regarding voice (VoiceWin()). Might help you.
`self.engine = pyttsx3.init("sapi5")
 voices = self.engine.getProperty("voices")
 self.engine.setProperty("voice", voices[1].id)
 self.engine.setProperty("rate", 140)`

Methenitis:
Hey! 

After communicating with @chekoduadarsh , my colleague @annakotsa and I have worked on adding GTTS as an option for voice generation and fixed the parameters of pyttsx3 in order to make the voice sound less robotic.

Would love to get your feedback on this @pnhofmann 

Thank you in advance!

pnhofmann:
#686

PR #686, opened 2020-05-12
Fix Jarvis Voice: Add GTTS engine & fix default engine (Issue #663)

Methenitis:
This PR works on issue #663 and adds the option of using the Google Text To Speech engine for the generation of Jarvis voice. The user is informed about data concerns with providing information to google and is asked for consent before GTTS is enabled. 

In case there is no consent given, Jarvis uses the engines already used for each OS. We have changed some parameters of the pyttsx3 engine used for windows, in order to make the voice, less robotic and more enjoyable. 

Changes made: 

- Added to plugins on `plugins/voice.py` that disable and enable GTTS accordingly. 

- Added gtts and playsound modules on requirements.txt in order for them to be downloaded on every build. 

- Added two functions on `CmdInterpreter.py`, that are called by the new plugins of Jarvis.py, in order to handle gtts

- Changed the displayed text when enabling sound on `plugins/voice.py`, to include the consent of GTTS engine

- Added a function to `utilities/voice.py` that gets as an argument the status of the gtts engine (enabled/disabled) and generates the correct voice engine accordingly

- Added a function to `utilities/voice.py` that generates text to speech using the gtts engines

- Updated a function to `utilities/voice.py` in order to fix the parameters of the voice windows use

- Fixed small PEP8 compliance issues in different files, as pointed by travis


pnhofmann:
Nice! Like really nice!

Just this commit is somehow strange (https://github.com/sukeesh/Jarvis/pull/686/commits/8734ff32fdfd1e28436d1f51efc7f5ca959f6e11). This files added should not exist; all tests should lie in ``jarviscli/tests``.

Yeah google does sounds good!

Just noticed, on Linux does require gstreamer + python gobject binding. Nothing you have to worry about, but maybe I'll find a workaround. Or I would just add a try-except ;).

Maybe we should also just swap two lines and make Jarvis first print out and then read out load.

But yeah, like this feature!

Methenitis:
Hi again! 

Really glad you like the feature! After the latest commits, most of the text beside the initial text would be printed before said out loud by Jarvis. Any ideas on how to get the initial text printed before Jarvis speaks? 

I was also confused about these initial commit files appearing when I first cloned the repo (talking about commit 8734ff3). What do you think I should do with these files? Just delete them and push again?

Let us know if we should do anything else on the feature before the merge. 

Thank you!

pnhofmann:
Hi!
> Just delete them

Yeah, did that for you ;).

Also switched from playaudio to pydub. Nothing wrong about playaudio - in fact pydub uses playaudio as one provider for playing audio. But pydub does support other audio playback providers like ffplay (ffmpeg) => that's better on Linux, since playaudio requires PyGObject (which requires native extension).

Also made small adjustment, that the intro is first print and then spoken.

So, everything fine for me; if you are OK about my changes, would merge.

Methenitis:
Hey @pnhofmann !

Thank you for your contribution, Pydub looks like the right choice and the print sequence change is great.

We are OK with your changes and would merge.

pnhofmann:
Thanks!

PR #689, opened 2020-05-13
Fix create_plugin.py + some various files to match with PEP8 guidelines

KonstantinosLolos:
Made some changes to the create_pluging.py file and some other files that made travis build fail because of PEP8 violations.  #674


pnhofmann:
Thanks ;)

Issue #676, opened 2020-04-10
Make bulkresizer plugin more user friendly.


gianniskokkosis:
Hello! I 've been looking through some code and I noticed that the bulkresizer plugin has room for improvements. More specifically:

- Jarvis' speech uses plain white letters (and they appear in the wrong order upon start)

- If the user gives an invalid path jarvis throws an unhandled exception and the user dosen't get the chance to give another path. 

- Could add more specific and detailed documentation

I would like to work on theese issues because I really like this project and I would love to contribute if it's ok with you. Let me know. Cheers!

pnhofmann:
Sure!

jay24rajput:
Hey is this issue up for grabs?

gianniskokkosis:
@jay24rajput Hello. I have finished the changes I wanted to make in bulkresizer plugin. In a couple of days I will make a PR. It's ok if you want to make some changes too!

jay24rajput:
Hey @gianniskokkosis sorry for barging in! Please go ahead with your PR. Cheers!

> @jay24rajput Hello. I have finished the changes I wanted to make in bulkresizer plugin. In a couple of days I will make a PR. It's ok if you want to make some changes too!

pnhofmann:
#692

PR #692, opened 2020-05-17
Make bulkresize plugin more user frienldly

gianniskokkosis:
Hello!
I have finished some changes on bulkresizer plugin. This PR works on issue #676 . Please review the changes and let me know if there are any problems or issues that I need to fix @pnhofmann .

Changes made:

- Add colors on Jarvis' speech in order to look nicer.

- Split the code to smaller and more specific functions. This change aims on making the functions of the code more specific in order to execute a certain functionality and not multiple functionalities. 

- Add functions that help the user to retype a path that leads to a valid dir, if she/he gives a path that lead to an invalid or not an existing dir

- Add a detailed documentation to all code's functions

- Add new capability to bulkresize plugin that creates an output path with the resized images if the path dosen't exist

P.S. I would like to add and some testing code as well! Tell me if it's ok with you.

### DIFF
--- a/jarviscli/plugins/bulkresize.py
+++ b/jarviscli/plugins/bulkresize.py
@@ -55,20 +55,20 @@ def spin(jarvis, s):
 
     """
     jarvis.say(' ')
-    jarvis.say('This is bulk resizer it will resize the image into given',
-               'size with padding !!!! Specially designed for Deep Leanring',
-               ' data collection process')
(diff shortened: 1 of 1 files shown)


pnhofmann:
Thanks! Excellent documentation ;).

> P.S. I would like to add and some testing code as well! Tell me if it's ok with you.

Sure! Testing is always a good idea ;).

PR #694, opened 2020-05-19
Add test cases for create_plugin.py and minor changes in create_plugin.py

KonstantinosLolos:
In this pull request:

- I have added some test cases for the create_plugin.py

- created an extra function to make the code cleaner.

- Fixed minor "too many blank lines issue in optional.py

I want also to add test cases for the plugin itself but I am having some trouble with mocking (I am now getting into it) and I cant seem to test the plugin without creating a file (as the plugin's utility is). Any tips as to where to look for info?
Thanks!
#674

### DIFF
--- a/installer/optional.py
+++ b/installer/optional.py
@@ -123,7 +123,6 @@ WKHTMLTOPDF = {
 }
 
 
-
 OPTIONAL_REQUIREMENTS = [PortAudio, RequestsSecurity, FFMPEG, ESPEAK, WKHTMLTOPDF]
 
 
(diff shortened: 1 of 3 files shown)


pnhofmann:
>  and I cant seem to test the plugin without creating a file 

There is in fact a mocking-module for file-operations! Docu with example code:

https://docs.python.org/3/library/unittest.mock.html#mock-open

Generally speaking, it is also possible to create a temporary directory (for example using https://docs.python.org/3/library/tempfile.html) and running test-code in this there.
But in your case; the mocking-solution is probably just more "elegant" ;).

KonstantinosLolos:
Decided to go with mocking the os.System method so the file is not being created every time the tests run. What do you think?

pnhofmann:
Nice ;)

Issue #696, opened 2020-05-24
queue_input method works more like a stack rather than a queue


KonstantinosLolos:
Hi!
While working on some testing code with my colleague @gianniskokkosis We discovered that the queue_input method for plugin testing works in the logic of LIFO and thought that it would probably more reasonable to be called stack_input or something like that (+ it would clear out the logic of it a bit more).
Wondering if you have any idea on how this could change without having to replace everything manually.
It's something really minor but thought it could help in the comprehension of the method.
Opinion?
*not really an issue, more like a thought*

pnhofmann:
Nice catch ;).

I think, just like the name suggests, FIFO-Queue is probably what you'd expect of such a method. And what I had in mind.
So yes, this is a bug.
A bug that is unnoticed, because there just isn't a test, where order makes any difference. Let's fix the behavior of queue_input.

KonstantinosLolos:
Great! Me and @gianniskokkosis can work on it! Any pointers and suggestions are more than welcome! :)

pnhofmann:
#697

PR #697, opened 2020-05-25
Fix bug of _input_queue attribute acting like a stack

KonstantinosLolos:
Fix bug of _input_queue attribute in function queue_input at __init__.py acting like a stack
Utilizes  the deque module from collections library to change the
behaviour from LIFO to FIFO as it should #696


pnhofmann:
Thanks!

PR #698, opened 2020-05-25
Added mass conversion functionality

shoniavika:
Based on length/time conversion functionality, added mass conversion one with unit tests.


pnhofmann:
Good work! Like this plugin - really useful. Just recently had to convert some oz -> g in a data sheet - now I can ask Jarvis ;). 
And with tests - perfect!

Thanks!

PR #707, opened 2020-06-13
Create a new Alexa skill for Jarvis

annakotsa:
Hello! This is the new Jarvis Alexa skill we have created along with my colleague @Methenitis. 
It includes the features of Spin The Wheel (Game), Calories (Health), Temperature Converter (Conversion), Stock (Finance), and Gmail (Communication). The reason why we chose to implement these features in our skill is that we wanted to include an indicative feature of as many Jarvis features categories as possible. 
We have also included the Jinja templates file and a README document with instructions in case someone wishes to add more things to the skill or even create new Alexa skills from scratch. 
Cheers, and thank you in advance!


pnhofmann:
Thanks!
Merged into new branch 'alexa-dev' as a good base for further development.

PR #710, opened 2020-06-14
Created plugin to get U.S. state capitals and postal abbreviations

arjunk2004:
So, for this plugin, I couldn't find a good API. Instead, I just ended up using two hard-coded dictionaries. This is probably a bad idea, so please let me know if you know of a good API for the task.


pnhofmann:
Thanks!

Issue #703, opened 2020-06-10
Non-interactive mode for Jarvis


arjunk2004:
Hi, I'm fairly new to this project. I was wondering if there was a non-interactive mode for jarvis. Something like where you could just go `jarvis "[command]"` and get the normal output.

arjunk2004:
I’d like to try to fix this if you don’t mind @pnhofmann

pnhofmann:
Sure!

arjunk2004:
Okay, I created a pull request (PR #711). The only catch is: for existing versions of jarvis, you need to delete the main jarvis file. Then, you should rerun the setup.sh script and make sure to do these steps in the processes (in the "Write Jarvis starter" part of the script):
  1) Install jarvis /usr/local/bin starter (requires root)
  2) Add /home/qbp/projects/Jarvis to $PATH (.bashrc)

I'm going ahead and closing the issue because I think it's all fixed, but please let me know if there are any problems!

PR #711, opened 2020-06-14
Passed command in launcher, so that non-interactive mode works

arjunk2004:
You may need to delete the jarvis file and rerun setup.sh for this to work on existing versions of jarvis.
Fixes issue #703


pnhofmann:
Will do the job, Thanks!

PR #712, opened 2020-06-14
Added plugin numbersapi

gdara17:
User inputs numbers, plugin displays facts provided by http://numbersapi.com/

### DIFF
--- a/jarviscli/plugins/numbersapi.py
+++ b/jarviscli/plugins/numbersapi.py
@@ -1,0 +1,34 @@
+import requests
+from colorama import Fore
+from plugin import plugin, require
+
+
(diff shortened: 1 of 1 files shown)


pnhofmann:
Nice and fun! Thanks!

As a minor suggestion, you could check that number is an actual number to avoid something like:

```
~> What can I do for you?
numbersapi asdf
	<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /asdf</pre>
</body>
</html>
```

pnhofmann:
Thanks!

PR #714, opened 2020-06-15
Add support for Windows 10 notifications

aprzn123:
I've added support for Windows 10 notifications using [win10toast](https://github.com/jithurjacob/Windows-10-Toast-Notifications). I've also created two potential icons for normal and critical locations, although they might need to be changed.


pnhofmann:
Thanks!!!

PR #715, opened 2020-06-17
A plugin for trivia questions

arjunk2004:
Let me know if there are any problems!


pnhofmann:
Nice!

But for multiple choice, I (as a user) would really like to know, what the correct answer was - instead of stay unknowing ;)

arjunk2004:
That's a good point, I'll push a fix for that soon. Shouldn't be too complicated

arjunk2004:
It should work now @pnhofmann

pnhofmann:
Thanks!

PR #725, opened 2020-07-01
Fixes #724 (path with spaces)

### LINKED ISSUE #724: BUG: Laucher doesn't work if the path to Jarvis directory contains spaces

alexandr7035:
Now spaces can be used in Jarvis dir's path


pnhofmann:
Thanks!

PR #730, opened 2020-07-10
added upside-down plugin

enqidu:
Hello, I added upside-down plugin that rotates the text, for example:
"chicken" will be ---> uǝʞɔᴉɥɔ


pnhofmann:
This is really nice! Thanks!

PR #731, opened 2020-07-11
Lang detection

sprintyaf:
I added language detection based on [fasttext's language classifier](https://fasttext.cc/docs/en/language-identification.html).
It can support 176 languages. 
I added two files in the data directory: the first is the model itself and the second is a dictionary between language codes and corresponding languages (because model returns language codes).

Example
example


pnhofmann:
Looks really good!

Didn't knew fastText, this is really fascinating what interesting stuff I get to know. Just noticed two smaller issues:

* This file ``jarviscli/data/lid.176.ftz``. It is MIT-licenced like complete fastText, so no big license issue. However we have to somehow stat (best in ``jarviscli/data/LICENSE`` that this is not our file and Facebooks owns copyright (but did license under MIT). So basically copy-paste [original license](https://github.com/facebookresearch/fastText/blob/master/LICENSE) and above say 'this applys to lid.176.ftz'.

* fastText requires native C++ extension. This is not a deal-breaker, but I'll update installer to handle fastText as an optional requirements since we like core-jarvis to run with pure-python.

sprintyaf:
> Looks really good!
> 
> Didn't knew fastText, this is really fascinating what interesting stuff I get to know. Just noticed two smaller issues:
> 
> * This file `jarviscli/data/lid.176.ftz`. It is MIT-licenced like complete fastText, so no big license issue. However we have to somehow stat (best in `jarviscli/data/LICENSE` that this is not our file and Facebooks owns copyright (but did license under MIT). So basically copy-paste [original license](https://github.com/facebookresearch/fastText/blob/master/LICENSE) and above say 'this applys to lid.176.ftz'.
> * fastText requires native C++ extension. This is not a deal-breaker, but I'll update installer to handle fastText as an optional requirements since we like core-jarvis to run with pure-python.

Thanks for your support. I added the license.

pnhofmann:
Made Fasttext an optional requirement - ready to merge!

Really nice feature, think we can even use this code to improve translate. 
https://github.com/sukeesh/Jarvis/blob/master/jarviscli/plugins/translate.py

PR #732, opened 2020-07-11
Tasks plugin

dudupopkhadze:
Hello @pnhofmann :), i added new tasks plugin, which helps user with tasks managment, i thinks it's pretty useful feature

With Tests of course :), Check it out


pnhofmann:
Nice!

Thanks!

PR #733, opened 2020-07-11
[NEW PLUGINS] Word Meanings, Synonyms and Antonyms

sprintyaf:
I added three new plugins. 
- "synonyms of" returns synonyms of a given word
- "antonyms of" returns antonyms of a given word
- "meaning of" returns different meanings of a given word

example
ex

### DIFF
--- a/installer/requirements.txt
+++ b/installer/requirements.txt
@@ -38,3 +38,4 @@ windows-curses; sys_platform == 'win32'
 Image
 img2pdf
 pdf2image
+PyDictionary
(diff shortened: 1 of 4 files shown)


sprintyaf:
This is what happens when arguments are unknown words
ex2

pnhofmann:
Sure!
Provides similar features like ``dictionary`` - but dictionary uses limited offline data, this is online. So this is additional value.

I'll merge - just let me thinks a bit, if we can integrate your new dictionary-plugin with old offline-dictionary plugin ;).

And to copy-paste small hint:

A small link:
https://www.python.org/dev/peps/pep-0008/

Pep8 are various best-practises how to format python code. You can check pep8 for example with flake8
flake8 --select E,W --max-line-length=140 --ignore E722,W503,W504,E128 jarviscli/ installer

Also tools like autopep8 do the job automatically formatting everything according to PEP8. Just as a hint for future PRs ;).

PR #735, opened 2020-07-12
Caesar cipher plugin added

shotanozadze:
This new plugin converts text into encrypted one (caesar cipher)

### DIFF
--- a/jarviscli/plugins/caesar_cipher.py
+++ b/jarviscli/plugins/caesar_cipher.py
@@ -1,5 +1,6 @@
 from plugin import plugin
 from colorama import Fore
+import string
 
 
 @plugin("caesar cipher")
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/caesar_cipher.py
```diff
@@ -0,0 +1,84 @@
...
+
+    jarvis.say(converted, Fore.YELLOW)
+
+
+def cipher_to_plain(jarvis):
+    user_input = get_user_input(jarvis)
+    converted = ""
+
+    for i in user_input:
+        if i.isalpha():
```
pnhofmann, line 60: Good work!

Only noticed, you did not consider Unicode. Something like for example our German: ä, ü, ö, ß ;). As a suggestion, you could instead isalpha do a check similar to:
https://stackoverflow.com/questions/196345/how-to-check-if-a-string-in-python-is-in-ascii#answer-196392

shotanozadze: Ok, I'll fix it

shotanozadze:
Done! @pnhofmann

pnhofmann:
Thanks!

PR #736, opened 2020-07-12
Near earth objects plugin added

shotanozadze:
This plugin helps us to find out which objects are near to earth at the given date and detailed information about them (api from NASA)


pnhofmann:
NASA - that is interesting. Thanks!

pnhofmann:
Hey, this NASA API also provides quite some interesting photos from earth, mars rovers etc. Currently thinking, maybe we could add even more features like ``show earth``?

shotanozadze:
NASA has very interesting APIs, it is good idea to add more features.
I'll try to add more. Let's see.

PR #742, opened 2020-07-14
Roulette game plugin added

shotanozadze:


pnhofmann:
Brilliant!

``flake8 --select E,W --max-line-length=140 --ignore E722,W503,W504,E128 jarviscli/ installer`` reports multiple missing whitespaces and ``comparison to True should be 'if cond is True:' or 'if cond:'``, but that is just syntax.

Readable code, did not find any issue; year good work! Thanks for this nice plugin!

As a suggestion - but really only a **suggestion**: What would you think about adding some kind of spinning-animation?
Like similar to: https://github.com/sukeesh/Jarvis/blob/master/jarviscli/utilities/animations.py#L21-L26
But maybe with numbers of a roulette number plate. Probably something like (completely untested):

```
number_plate = [0, 32, 15, ...]
ball_index = 0
ball_speed = random(10, 15)

while True:
    # print over old number
    sys.stdout.write('\r' + number_plate[ball_index])
    sys.stdout.flush()

    # move ball
    ball_index += ball_speed

    # randomly slow down ball
    if random.randint(0, 10) > 3:
        ball_speed -= 1
    if ball_speed == 0:
        break
    sleep(0.2)

```

maybe even with colors red / black?

shotanozadze:
Thanks for your support.
I'll think a bit about your suggestions to decide what to add.

pnhofmann:
I'll just merge, in case you / anyone who read this has new ideas, just open a new issue / PR ;).

Thanks!

Issue #746, opened 2020-07-20
Curl Request Generator


adarshtri:
A plugin that could generate curl request for the users based on the HTTP method, content type, data.

PR #747, opened 2020-07-21
Plugin to generate curl request

adarshtri:
#746   @pnhofmann  a new plugin to generate curl request based on collecting inputs like HTTP method, endpoint, content-type, data from the user. The plugin has basic validators for user input.


pnhofmann:
Hi and Thanks!

This is a really useful plugin. I need curl quite regular for trying some web apis ;).
And well structured, code.

PR #752, opened 2020-07-26
expression validation checker implementation

vvash17:
implementation of expression checker : 
checks if arithmetic expression is valid. 
for example :         
                    1) (), )),((,)(, )4,5( ... AND SO ON NOT VALID!
                    2) (5+3), (5+2)*(3/2) VALID


pnhofmann:
Nice! Really think we should use this code also for validating expressions in our maths plugin ;).

PR #756, opened 2020-08-01
Update e_launcher.py

dbherol:
The original file had JARVIS_MACRO assigned to the string:
```
    JARVIS_MACRO = """\
    #!/bin/bash
    source "{PATH}/env/bin/activate"
    python "{PATH}/jarviscli" "$@"
    """
```

This string creates indents before each line, causing the "jarvis" executable file to be unreadable on mac os.

Results in the error: 
```

Failed to execute process './jarvis'. Reason:
exec: Exec format error
The file './jarvis' is marked as an executable but could not be run by the operating system.
```

A simple fix to remove the indents is to change the JARVIS_MACRO assignment to:
```
    JARVIS_MACRO = """\
#!/bin/bash
source "{PATH}/env/bin/activate"
python "{PATH}/jarviscli" "$@"
    """
```

### DIFF
--- a/installer/requirements.txt
+++ b/installer/requirements.txt
@@ -31,6 +31,7 @@ sympy
 tabulate
 termdown
 opencv-python
+whois
 wikipedia
 win10toast; sys_platform == 'win32'
 gtts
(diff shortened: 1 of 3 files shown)


pnhofmann:
Thanks!

Didn't knew there is a difference with leading spaces Linux <-> MacOS - Linux these indents made no issue, and therefore didn't noticed.

Also thanks for get_host_info.py plugin! But it's better to send separate Pull Requests for different subjects ;). Nevertheless, really nice plugin!

PR #769, opened 2020-10-01
Added new plugin - Jokes

JeswinSunsi:
This PR adds a new plugin to Jarvis. It tells a random joke from the Pyjokes module. Network connection is necessary.


pnhofmann:
thanks :)

PR #776, opened 2020-10-01
Update README.md

mayank1897:


mayank1897:
try to 
import pyttsx3
for text-to-speech transformations

PR #779, opened 2020-10-01
added offline speech recognition for voice_control.py and added internet control plugin

erick-dsnk:
Added offline speech recognition using CMU Sphinx, check [voice_control.py](jarviscli/plugins/voice_control.py), fixed a few typos and added a new internet control plugin, see [internet_control.py](jarviscli/plugins/internet_control.py)


jarviscli/plugins/voice_control.py
```diff
@@ -1,5 +1,7 @@
 import os
-from plugin import plugin, require
+from jarviscli.plugin import plugin, require
```
pnhofmann, line 2: Why this change? Importing jarviscli.plugin actually works for you?

jarviscli/plugins/voice_control.py
```diff
@@ -15,27 +17,50 @@
...
                 print("Say listen to start listening")
                 r.adjust_for_ambient_noise(source)  # Eleminating the noise.
                 audio = r.listen(source)  # Storing audio.
-                pinger = r.recognize_google(audio)  # Converting speech to text
+
+                if connected:
+                    pinger = r.recognize_google(audio)  # Converting speech to text using google recognition.
+                
+                else:
+                    pinger = r.recognize_sphinx(audio)  # Converting speech to text using Sphinx CMU in case user is not connected to internet
```
pnhofmann, line 56: Probably needs a check if sphinx (pip package: pocketsphinx) is installed. And we should add pocketsphinx as an optional requirement to our installer.

jarviscli/plugins/voice_control.py
```diff
@@ -44,18 +69,23 @@ def hear(jarvis, s):
...
+                    pinger = r.recognize_google(audio).lower()
+                
+                else:
+                    pinger = r.recognize_sphinx(audio).lower()
 
             if (pinger == "stop"):
                 listen = False
                 print("Listening stopped.")
                 _jarvis.speech.text_to_speech("Listening stopped.")
-                break
```
pnhofmann, line 58: Why remove this break?

jarviscli/plugins/voice_control.py
```diff
@@ -15,27 +17,50 @@
...
+                
+                else:
+                    pinger = r.recognize_sphinx(audio)  # Converting speech to text using Sphinx CMU in case user is not connected to internet
+
             try:
                 if (pinger.lower() == "listen"):
                     listen = True
                     _jarvis.speech.text_to_speech("Voice mode activated")
                     print("Voice mode activated. Say something!")
-                    break
```
pnhofmann, line 38: Why remove this break?

jarviscli/plugins/voice_control.py
```diff
@@ -15,27 +17,50 @@
         'voice_control_requirements (install portaudio + re-run setup.sh)']
 
 
+def has_internet():
```
pnhofmann, line 20: +1
We could probably even move this function to JarvisAPI so it's available to all plugins.
And/or use this to (finally) implement ``@require(network=True)`` ;).

jarviscli/plugins/voice_control.py
erick-dsnk: yes it does, and i was getting errors from vs code as well about not finding the file and couldnt get the suggestions. but yes, it works for me in this format

jarviscli/plugins/voice_control.py
erick-dsnk: i had a bit of a pain installing pocketsphinx with pip. im not exactly sure why, but installing directly from a wheel worked for me. 

thats definitely something I'll look into though.

jarviscli/plugins/voice_control.py
erick-dsnk: the while loop is only active as long as `listen = True`. In that code block you're setting listen to False therefore it will break the loop, so there is no point in that break

jarviscli/plugins/voice_control.py
erick-dsnk: yep, that sounds like a great idea :)

erick-dsnk: @pnhofmann can you guide me into where the plugin requirements are checked? I want to help implement the network requirements,  but I'm not sure where to start. My best guess was in plugin.py in `_plugin_run_with_network_error` because the docstrings said that the method is auto run when `require(network=True)` is provided.

pnhofmann:
Hi again!

Sorry, it's a bit complicated since I made some changes on similar parts, then wanted to finish this changed, then postponed everything and sort of forgot you ;).

So, finally: I'll just merge your PR in my dev branch, add some more voice control improvements (I wanted to add for quite a while) and then everything will be merged into master (hopefully soon ;) ).

Thanks by the way for these commits ;).

> can you guide me into where the plugin requirements are checked?

Yep, ``jarviscli/plugin_manager.py`` - there's a class ``PluginDependency``.

PR #786, opened 2020-10-02
Update Jarvis.py

KeshavGarg144:
A new file is added which contains some basic functions such as opening of kubernetees, calender , etc which is user voice enabled.


KeshavGarg144:
new.py was imported to enable voice enabled opening of apps.

appi147:
@KeshavGarg144 Reopen when and if you update.

PR #788, opened 2020-10-03
Mac Address Changer 

mikek4233:
FOR LINUX ONLY

A plugin for viewing and changing any devices MAC address connected to your computer

### DIFF
--- a/jarviscli/plugins/change_mac.py
+++ b/jarviscli/plugins/change_mac.py
@@ -1,16 +1,16 @@
-from plugin import LINUX, UNIX, MACOS, WINDOWS, plugin, require
+from plugin import LINUX, plugin, require
 from platform import system as sys
 from colorama import Fore
 import subprocess
 import re
-# import winreg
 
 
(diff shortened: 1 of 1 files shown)


mikek4233:
Not sure why this is failing though it looks like most recent PR are failing

pnhofmann:
> Not sure why this is failing though it looks like most recent PR are failing

Dont' worry ;).

Anyway, Looks good!

Thing is, 'ifconfig' is deprecated on most linux distributions, more and more are switching to 'ip'
https://www.redhat.com/sysadmin/ifconfig-vs-ip

Which wouldn't stop me from merging your PR, but your decision:
* If you like, you could take a look at ``ip`` to make your plugin compatible
* Or we could just add a ``@require(native=["ifconfig"])`` to mark this plugin only compatible to systems where ifconfig is installed.

mikek4233:
@pnhofmann I updated to use ip, thanks for the advice

pnhofmann:
Thanks!

PR #795, opened 2020-10-13
Wifi password getter

### LINKED ISSUE #793: WIFI Password Getter
This plugin will find and display all the profiles of the wifis that you have connected to and then display the password if selected 

Happy to work on this

mikek4233:
For Linux only

Returns a list of wifi's stored on your device and retrieves password based on selection

closes #793


pnhofmann:
Thanks!

PR #805, opened 2020-11-21
Adding a new Voter Information Plugin 

abcumm:
I added a Voter information plugin that gives the user information about where to vote as well as some information about what will be on the ballot. The info is provided by Google Civic Information API, which is currently set up to my API key. I also have an additional requirement in adding geopy which can turn lat/long coordinates into street addresses and is available through python pip. Let me know your policy on adding to your requirements and if there are any issues with using my API key as well as any other critiques you might have. Also I just want to say I love this project, it does an amazing job of being accessible to programmers new to open source like myself!


pnhofmann:
Nice!

Mental note @me: Take a closer look at geopy and maybe integrate in Jarvis core - plugins like ``weather`` could really use some improvements ;).

> which is currently set up to my API key

API keys and Open Source projects = ugly!

Obviously make sure, the API key is configured OK, so you won't be charged anything! If you and Google (maybe view Terms of Service?) are OK with usage of this API key, we are, too. 

And / Or you could ask the user to provide his/her own API key:
https://github.com/sukeesh/Jarvis/blob/master/jarviscli/packages/timeIn.py

For something like Google API key, this would be quite useful since - if I'm correct - this API key can be used for quite a lot APIs?

abcumm:
I used a similar process to the timeIn plugin to have the user provide their own api.

abcumm:
Yes it could be used with all of google's apis outside of maps, the key I had previously was restricted to just the Civic Information one but this way the users key could be used for other plugins in the future.

pnhofmann:
>  but this way the users key could be used for other plugins in the future

+1

Thanks!

pnhofmann:
PS: Maybe reset your google API key since it still appears in git history ;).

PR #806, opened 2020-11-27
New Feature: hex plugin that converts an integer to a hexadecimal number

izhang05:


pnhofmann:
Simple but useful plugin with test case - I like it! Thanks!

PR #812, opened 2020-12-20
[New Features] two new plugins , a speed converter and a morse code translator

siderishub:
I added a speed converter at speedconv.py (use command jarvis speedconv to use it) and
a morse code translator on morse_code.py (use command jarvis morsecode to use it) which has a morse code encoder and decoder


pnhofmann:
Nice! Nothing to add ;).

Thanks!

PR #816, opened 2021-01-21
Plugin: scan_network with nmap

metodiistatkov:
Hello,

I created a plugin that uses nmap to scan a network address. It could be helpful for a quick check of what devices are active in your LAN. Also, you can provide a specific address to be scanned - either private or public IPv4 one. As a precaution I have added the *Linux* requirement since I have tested it only under Linux. However, I suppose it should work fine on other OS as well, but will verify that in subsequent improvements. Note that nmap needs to be installed already in order to use this plugin.
Here are two examples of how the new Jarvis command works:

*This image demonstrates scanning a particular address*
Screenshot from 2021-01-21 16-17-26

*This example shows how you can scan your local network for example*
Screenshot from 2021-01-21 16-22-40

**Future Work**
I am planning to add more functionalities that give greater flexibility. For example, scan for a specific port range, check if a particular host is up, etc. Also I will add more test cases.


jarviscli/plugins/scan_network.py
```diff
@@ -0,0 +1,51 @@
...
+    Scans provided network for all connected devices
+    Usage example: jarvis scan_network 172.16.1.10/24
+    """
+    nm_scan = nmap.PortScanner()
+    jarvis.say('Please wait, this might take a while!')
+    if not s:
+        jarvis.say('You need to provide address as input!\n'
+                   ' For example: jarvis scan_network x.x.x.x/24', Fore.YELLOW)
+    else:
+        nm_scan.scan(s)
```
pnhofmann, line 43: Noticed python-nmap default adds '-sV' as option. According to man

``-sV: Probe open ports to determine service/version info``

This takes ages and in line 49 you are dropping this information anyway. As a suggestion, overwrite this parameter for extremely faster scan times:
```suggestion
        nm_scan.scan(s, arguments="")
```

jarviscli/plugins/scan_network.py
```diff
@@ -0,0 +1,51 @@
...
+        if 'error' in nm_scan.scaninfo():
+            error_output(jarvis, nm_scan)
+        else:
```
pnhofmann, line 46: Looks like nmap may still produce meaningful output even if an error was reported. At last on my system it reports:

``warning: Unable to determine any DNS servers.``

(don't know why) but still works. Maybe remove the else-clause and call the code in any case?

pnhofmann (approved):
Thanks!

Really good; PEP8, with test-case and a really useful feature!

> Future Work

Nice ;).

You decide: Shall I merge now and you create new PR - or keep this one open and you just push?

jarviscli/plugins/scan_network.py
metodiistatkov: Thanks for pointing that out!

metodiistatkov:
Thanks for the fast review!

Up to you really, I don't mind either way. If you want to have the functionality sooner probably merge it and I will implement your suggestions over the weekend and make a new PR.
 Also, I wanted to ask you about using Python type hints. I usually use type hints with python3, but I noticed that there are not any in the code. Is that because we want to have compatibility with python2? Should I add type hints in my next commits or refrain for doing that for now?

Btw, awesome project and idea!

jarviscli/plugins/scan_network.py
metodiistatkov: Makes sense, worst case scenario users will simply end up with larger error message.

pnhofmann:
Yeah, Jarvis is now Python 3 only - feel free to use type hints. I should probably use them more often, too ;).

jarviscli/plugins/scan_network.py
metodiistatkov: So I checked this in detail and although removing the arguments provides a significant speedup, there is a miss on important information. While I do remove the services, version info provides helpful information about the underlying operating system. Unfortunately, I could find a way to extract the version only - without the services; thus I am going to leave the default arguments and just place a comment providing your suggestion for people who would favor speed over detail depth.

metodiistatkov:
@pnhofmann ready to merge

pnhofmann:
Thanks!

PR #817, opened 2021-01-31
Update matrix_add.py

Akshit1510:
removing error if input string cannot be converted to string by using try and except block , replacing wrong input with 0


pnhofmann:
Interesting with try-except-else - don't this this one often.

Yeah, Thanks for fixing this one!

PR #820, opened 2021-03-01
Update README.md, Add poetry.lock file

Devansh3712:
- Add `python` formatting in README markdown
- Add `pyproject.toml` and `poetry.lock` file for dependency management


pnhofmann:
Hi!

> Add pyproject.toml and poetry.lock file for dependency management

Actually don't see any advantages for these files - but honestly don't really know much about Python Poetry. So that is really nothing against this tool, but I feel like requirements.txt is sufficient and easy enough for this project.

Would merge wrong language-tag in Readme, though (thanks for noting).

PR #827, opened 2021-03-19
Spotify feature

Pierrecoulon1:


jarviscli/plugins/spotify.py
```diff
@@ -0,0 +1,148 @@
+from plugin import plugin, require
+from colorama import Fore
+
+import sys
+import spotipy
+import spotipy.util as util
+
+SPOTIPY_CLIENT_ID = '865dcb503fc8489d9a2366b5fc0dc58d'
+SPOTIPY_CLIENT_SECRET = '898fbca894d6401785d5f061771fa25d'
```
pnhofmann, line 9: > Client Secret is the key that you pass in secure calls to the Spotify Accounts and Web API services. Always store the client secret key securely; never reveal it publicly! 

(https://developer.spotify.com/documentation/general/guides/app-settings/)

Please reset your key! It is acceptable to ask the user to provide his/her own key. For example like it is done in this emacs plugin:

https://github.com/danielfm/spotify.el

Here is an example of an user-provided key in this project:

https://github.com/sukeesh/Jarvis/blob/master/jarviscli/packages/timeIn.py

Probably it wouldn't be a bad idea to build some kind of API-management feature in Jarvis (I'll think about it).

evamy: @pnhofmann 
#859 maybe a key vault can help ?

PR #831, opened 2021-03-26
FIX newyear.py

Rahimi-Alireza:
new-year plugin was bugged and it didn't work so i changed declaration and it worked


pnhofmann:
Thanks

PR #832, opened 2021-03-26
Renamed files

Rahimi-Alireza:
I noticed files where using CamelCase so i changed to PEP8's underscore

*there is a mistake with new-year-fix and it is from my previous PR


pnhofmann:
Thanks!

PR #838, opened 2021-04-20
Fix directions_to.py and add an example

NikolasMoatsos:
While using Jarvis I found a bug in the "directions" command:

If the user enters: "directions from X to Y", the  variable "data" given in the main() function of directions_to.py is  a String in the form: "from X to Y", due to the strip() method. So, the if statement: `if  " from " in data:` is never satisfied, due to the trailling space and the user doesn't get the directions from the start city he wants.

This PR:

- Fixes the previously mentioned bug.
- Adds docstrings to the main() function of directions_to.py.
- Removes the wordIndex() function, which executes an already executed command and embeds its utility in the main() function, since the existance of the wordIndex() is redundant .
- Adds an example of directions with both start and destination city in converted.py.

### DIFF
--- a/jarviscli/packages/directions_to.py
+++ b/jarviscli/packages/directions_to.py
@@ -1,16 +1,21 @@
 from . import mapps
 
 
-def wordIndex(data, word):
-    wordList = data.split()
-    return wordList.index(word)
-
-
 def main(data):
+    """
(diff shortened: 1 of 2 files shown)


pnhofmann:
Good fix! Thanks!

PR #844, opened 2021-05-01
Fix small bugs in "near" command

NikolasMoatsos:
In the "near" command there are two small bugs, specifically :

1. If the user asks for: "things near city", the URL to be opened in the web browser is missing in mapps.py and an error occurs.
2. The if statement: `if ' me' in data`, wrongly interprets some arguments as "me" for example if the user asks, "restaurants near mexico city".

This PR : 

- Fixes the previously mentioned bugs.

Similarly to #838 :

- Adds docstrings to near_me.py.
- Removes the wordIndex() function and embeds its utility in the main() function, since the existence of wordIndex() is redundant.


pnhofmann:
Thanks!

PR #840, opened 2021-04-29
File Manager Plugin Created

Tpool1:
The plugin has 3 main functionalities: delete, move, and rename given the path for a file. I am willing to change anything needed and please make suggestions for other features, improvements, etc.

PR #846, opened 2021-05-05
Add tests and fix build

NikolasMoatsos:
This PR aims to : 

1. Fix the building failure in Travis CI.
2. Add some tests for near_me and directions_to.

## Fix building failure

To fix the building failure in Travis CI I made some changes, specifically : 

### For Linux and macOS :

- Add the test.py file in the custom directory, which is needed for the test_create_plugin.py and was deleted by #840.
- Fix flake8 errors by fixing some styling issues.
- Fix test errors by changing some false imports.

### For Windows : 

- Add the pip upgrade command in the .travis.yml file, in order to fix the cryptography install error.
- Fix the error occurring only in Windows, in test_bulkresize.py when importing the cv2 package :
image

After searching for this issue one feasible solution that I found was to change the opencv-python package in the requirements.txt with the opencv-python-headless version, which is the same with the standard opencv-python package but without the GUI functionality and since we do not use any GUI functionality in the Bulkresizer plugin where the cv2 package is imported, I believe this solves the problem.

## Add tests

After fixing some small bugs in the near_me.py and directions_to.py  with #844 and #838, I also added some unit tests to check their methods more accurately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ jobs:
       language: shell
       before_install:
         - choco install python --version 3.7.4
+        - python -m pip install --upgrade pip
       env: PATH=/c/Python37:/c/Python37/Scripts:$PATH
 install:
     - if [ "$TRAVIS_OS_NAME" == "osx" ]; then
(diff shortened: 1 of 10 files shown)


pnhofmann:
Big thanks!

PR #849, opened 2021-05-08
Fix clear plugin for Windows

nikiboura:
This PR aims to **fix the clear plugin** in order to work for many Operating Systems.

More specifically, there was a bug in the "clear" plugin as it didn't work for Windows, although it appeared in the --help-- plugin list.

After reviewing it, I added the proper command to correct the functionality.

I also added the needed documentation and formatted the code according to PEP8 instructions.

### DIFF
--- a/jarviscli/plugins/clear.py
+++ b/jarviscli/plugins/clear.py
@@ -1,15 +1,19 @@
 """
 Jarvis plugin for clearing the terminal
 """
-from plugin import plugin
+from plugin import plugin, require, WINDOWS, UNIX
 import os
 
 
+@require(platform=UNIX)
 @plugin("clear")
-def clear(jarvis, s):
-    if os.name == "posix":
-        # Unix/Linux/MacOS/BSD/etc
-        os.system('clear')
-    elif os.name in ("nt", "dos", "ce"):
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/clear.py
```diff
@@ -1,8 +1,15 @@
...
 
-@plugin('clear')
+@plugin("clear")
 def clear(jarvis, s):
-    """Clears terminal"""
-    os.system("clear")
+    if os.name == "posix":
+        # Unix/Linux/MacOS/BSD/etc
+        os.system('clear')
+    elif os.name in ("nt", "dos", "ce"):
```
evamy, line 13: Jarvis does the check for OS very early on, you can also use that instead of re-evaluating here.

pnhofmann:
> Jarvis does the check for OS very early on, you can also use that instead of re-evaluating here.

Yeah, pushed a commit for this. Could argue, if this actually makes the code cleaner but in any case:  Evaluating only once is a good idea, because if the OS-check does not do what it should, there is only one place to look for errors.

In any case thanks @nikiboura for showing us this issue and opening this PR.

PR #811, opened 2020-12-17
Marketplace feature! Two new plugins "market find" and "market buy"!

Martins6:
As discussed in the #801 the idea for a marketplace would be a nice addition to the Jarvis community. I've built two new plugins which allow the user to list features ("market find") and also to install new plugins from Github repositories ("market buy"). All the codes should be spotted in the Github Topics and follow the main Jarvis branch. All the documentation and proposal is written in the PLUGINS_MARKETPLACE.md.

An example of basket of plugin is found [here.]( https://github.com/Martins6/LinkedIn_Jarvis_Plugin ) (it is also an template repo.) This is a plugin to read your LinkedIn feed, it needs the Chrome browser and your credentials.

If something isn't clear in the plugins or in the general idea, please let me know! I shall try to explain better.

PS: Of course, there is lots of room for improvement!! Like how to install the necessary requirements without using the "python installer .", etc.

Issue #801, opened 2020-10-28
Library/Repository for Jarvis Plugins


Martins6:
I've built a Jarvis plugin to administer my R scripts, but I also know not everybody will be interested in this specific plugin. So, I was thinking:

1. How about we construct a repository or library for Jarvis plugins? Some server or someone's repository in Github where it would contain a bunch of plugins ready to be installed as needeed.

2. We should also build alongside an "install plugin" plugin, where it could install from this server/library/repository of codes! 

As it is fairly simple the "install plugin" plugin (simply extract file and copy it to jarviscli/plugins), I'd volunteer to build this script, however...for that we need the repository first. And that seems something to be discussed with the community.

Martins6:
Those jarvis plugins in the repository could be in a specific file type like: "example.jarvis_plugin". Many ways to do the same job!

pnhofmann:
Like a marketplace ;). Like this idea.

I'd also have no problem with plugins in this repo, which are disabled by default. But I agree, for highly specific plugins a own repository would be better.

My two cents:

> simply extract file and copy it to jarviscli/plugins

I'd simply ``git submodule``  the repositories to a ``jarviscli/additional`` folder (included in plugin search path).

Martins6:
I've got to be frank, didn't know what git submodule was. Then I searched for it and found it amazing! Nice git functionality. 

Anyways...like you said @pnhofmann, for the more accepted plugins the ` git submodule ` would work just fine. However...I'd argue that as Jarvis grows the 'marketplace' will become more and more necessary. So...why not start with that? It will solve for the more accepted plugins but It will likewise solve for the more specific ones too!

The difficult part, I'd wager, would be creating the infrastructure and the community rules.

pnhofmann:
>  git submodule would work just fine.

OK, need to rollback from this spontaneous idea ;). Didn't think this through: Git submodule means, this repo is added to git history and we certainly don't want to commit what plugins we installed.
What we'd need would be rather a regular ``git clone`` (like e.g. my vim plugin manager does it) - and obviously exclude ``jarviscli/additional`` from Jarvis git history.

> We could also elaborate a way to 'download' a bunch of plugins, like they are tagged together, at the same time. So there could be a canonical 'extra' plugins, the more accepted ones, that would be just a one line of code away. That way everyone is happy!

+1

> The difficult part, I'd wager, would be creating the infrastructure and the community rules.

Another suggestion - which isn't really from me, but I have seen similar solutions:

[...]

PR #851, opened 2021-05-09
Adding Zork Plugin

DobbyK:
Please note the actual game content is not mine, and I just ported it over for Jarvis. I will eventually expand the content (possibly), and may even make a MUD out of it if anyone would be interested.

Also should Include the License into the code itself, or keep it a separate file?

### DIFF
--- a/jarviscli/plugins/LICENSE.txt
+++ b/jarviscli/plugins/LICENSE.txt
@@ -1,0 +1,18 @@
+Copyright (c) 2014 Billy Burrows
+
+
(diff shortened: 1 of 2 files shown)


DobbyK:
Does anyone know why the checks failed?

pnhofmann:
> Please note the actual game content is not mine

Sorry, we only accept work that is licensed MIT or compatible.

There is the idea of a marketplace (-> #801 #811) for plugins not maintained by us. But it's not yet merged into master...

PR #855, opened 2021-05-13
Documentation Updated (#1)

DobbyK:
This PR, improves the documentation of the plugins, and I am still working on improving it. I am also working on my own plugin, for a online chat room for everyone to chat in, and I was wondering who would be able to host it, or if I would have to myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ jarvis is cool
 
 [Click here](doc/PLUGINS.md) to learn more about plugins.
 
+[Click here](doc/PLUGINS-LIST.md) for a list (not complete) of plugins that come with the master branch.
+
 ### Creating a test
 
 Creating a test is optional but never a bad idea ;).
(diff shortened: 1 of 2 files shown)


pnhofmann:
Can we auto-generate this file - since this would make it easier to maintain this file?

I really like it how you group plugins by functionality. Maybe we could even add something like ``@plugin(name=..., group=...)``

DobbyK:
Amazing Idea! I could try something similar to that.

DobbyK:
I assume I should add it in jarviscli/plugin.py

pnhofmann:
> I assume I should add it in jarviscli/plugin.py

Yep. With with decorators this code is probably not the most 'trivia' code - if you have any questions, please ask ;).

DobbyK:
I have a basic way of editing the file, but I'm not sure how to change to the specific line based on group, I can commit it if we want to get rid of the grouping.

pnhofmann:
Yes, commit and let me take a look ;).

PR #874, opened 2021-06-26
Marketplace enhancement and new documentation!

Martins6:
I made some refactoring in the code deleting what was useless, fixed the documentation in the doc/ section, and also made some new enhancements, like making the features OS agnostic!

### DIFF
--- a/jarviscli/plugins/market_buy.py
+++ b/jarviscli/plugins/market_buy.py
@@ -1,8 +1,4 @@
-# Modules
-# Installing plugins
 import os
-from pathlib import Path
-
 from plugin import Platform, plugin, require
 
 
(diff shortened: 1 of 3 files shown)


jarviscli/plugins/market_buy.py
```diff
@@ -68,6 +49,15 @@ def __call__(self, jarvis, s):
...
                 for f in files:
                     if f == 'requirements.txt':
-                        os.system('env/bin/pip install -U -r ' + os.path.join(root, f))
+                        jarvis_folder = os.path.split(self.MARKETPLACE_PATH)[0]
+                        os.system((f'cd {jarvis_folder} && ' +
+                                   'env/bin/pip install -U -r ' +
+                                   os.path.join(root, f)))
+
+        # Deleting the .git folder from the cloned repo.
+        shutil.rmtree(os.path.join(plugin_dir, '.git'))
```
pnhofmann, line 58: Why? Don't see benefit - doesn't this just make updates complicated?

Apart from that, good work! Really good idea to use python-git binding instead os.system ;).

jarviscli/plugins/market_buy.py
```diff
@@ -68,6 +49,15 @@ def __call__(self, jarvis, s):
...
+
+        if self.MAINTAIN_PYTHON_ONLY:
```
pnhofmann, line 60: Actually, don't see a benefit; something files like LICENSE or README don't waste any significant space and are nothing I would delete.

Martins6:
After a long time...I finished the refact of the code! And the projects keeps growing and that is so cool to see! 

Also, @pnhofmann, I've also created a more simple template for the marketplace in: https://github.com/Martins6/jarvis_marketplace_template

pnhofmann:
Thank you!

> And the projects keeps growing and that is so cool to see!

Yeah, I wanted to merge dev branch by the end of last year; now it's in the middle of march ;). But it'll happen eventually...

PR #890, opened 2021-08-17
Bug-Fix: First Reaction Logical Control

AhmetFurkann:
In the Jarvis API, the first_reaction boolean value has no effect.

As can be seen in the image, when the `first_reaction` is set to `False` value, it has no effect.

fix:first_reaction_control

So I have added a logical control for the boolean value of `first_reaction` in the constructor of class of  the `CmdInterpreter`.


pnhofmann:
Thanks ;)

AhmetFurkann:
You are welcome :)

Issue #862, opened 2021-06-03
[NEW FEATURE] Book Suggest


mrmuke:
I like to use selenium or bs4 to get top book suggestions based on what they've read or just something new.

pnhofmann:
Probably bs4 better than selenium, because selenium is a bit tricky to ensure to really work on all systems ;).

Web Scrapping - always a bit tricky. Just make sure, the owner of the website has no problem with what you plan to do ;).

26tanishabanik:
I have the added the above said feature, https://github.com/sukeesh/Jarvis/pull/892

mohitrbhardwaj:
I am new to open source. Would like to contribute here but don't know how to start. Can somebody help please?

PR #892, opened 2021-08-20
Added new feature Book Suggestions

26tanishabanik:
I have added a new feature of suggesting books using beautiful soup based on genres given by the user. Issue #862

### DIFF
--- a/jarviscli/plugins/books.py
+++ b/jarviscli/plugins/books.py
@@ -1,0 +1,78 @@
+from plugin import plugin, require
+from bs4 import BeautifulSoup
+import requests
+
+
+@require(network=True)
+@alias("books", "book")
+@plugin('book_sugges')
+def extract_genre(jarvis, s):
+    """
+        This method gives jarvis the ability to find a list of books
+        based on the user's interest of genres.
+    """
+    jarvis.say("Hey! Ready to turn some pages? \n")
+    jarvis.say("Here are some of the interesting genres: ")
+    genres = [
+        "romance",
+        "fiction",
+        "fantasy",
+        "science-fiction",
+        "non-fiction"
+        "",
+        "children",
+        "history",
+        "mystery",
+        "horror",
+        "historical-fiction",
+        "love",
+        "biography",
+        "thriller",
+        "contemporary",
+        "classics",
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/books.py
```diff
@@ -0,0 +1,78 @@
...
+        try:
+            book = item.find('a', class_='bookTitle').span.text.strip()
+            author = item.find(
+                'div', class_='authorName__container').a.span.text.strip()
+            rating = float(
+                item.find(
+                    'span',
+                    class_="minirating").text.strip().split(" ")[0])
+            record = (book, author, rating)
+        except BaseException:
```
pnhofmann, line 57: Looks good!

I'll take a closer look tomorrow. Just small note, catching all Exceptions is usually not a good idea. Might produce hard-to-find bugs ;).

26tanishabanik: @pnhofmann , I didn't actually catch all exceptions, it was converted automatically by the autopep8 package. It was just except earlier, so as to prevent from entering empty record to the final records list.

jarviscli/plugins/books.py
```diff
@@ -0,0 +1,78 @@
+from plugin import plugin, require
```
pnhofmann, line 1:
```suggestion
from plugin import plugin, require, alias
```

jarviscli/plugins/books.py
pnhofmann: Same thing if ``except Exceptions`` or just ``except``. Interesting autopep8 prefers the former.

It basically makes finding bugs harder since it hides all Exceptions. Imagine for example, for some reason ratings are measured like 4,5 (with a , like we do it here in Germany for some reason). With your code you will ask yourself, why is no result found instead getting a clean message "ValueError: could not convert string to float: '4,5'.

But please see it just as a general note. In this case the except probably isn't a real issue ;).

pnhofmann:
Thanks ;)

jarviscli/plugins/books.py
26tanishabanik: Yes, I understand, thanks

PR #893, opened 2021-08-23
Update requirements and added downloader for resources

AhmetFurkann:
### Requirements For snips-nlu
While try to implementing some feature, found some bug for required modules and resources.

Running the `python __main__.py` statement raises an error something like below.

Screenshot from 2021-08-23 13-50-53

That's why I have updated the requirements.txt file.

And also after installing the `snips_nlu` module, it is required the download a language resource. So `b_pip.py` has been updated for downloading the resources and added a txt file which is called `res_requirement.txt`. The error massage is like this:

Screenshot from 2021-08-23 13-54-48

### Requirements For Kivy
On the other hand for running the `python __main__.py --with-gui` requires the `kivy` framework.

Screenshot from 2021-08-23 14-00-36

The requirements.txt has been updated for `kivy`.


pnhofmann:
Hi! I really like how you handled snips download with the ``res_requirements.txt`` file! Thanks!

PR #898, opened 2021-08-28
New Feature (automation of installation of docker, Linux commands and Hadoop.

Ganesh-Chaudhari405:
This plugin is helpful for software Engineer. it helps them to automate installation of docker,Hadoop,basic linux commands,webserver configuration.

### DIFF
--- a/jarviscli/plugins/automation.py
+++ b/jarviscli/plugins/automation.py
@@ -2,7 +2,7 @@ import os
 from plugin import plugin, require, LINUX
 
 """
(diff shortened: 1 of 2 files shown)


pnhofmann:
Hi!
Hmm, does only work on Fedora / Red Hat... Maybe a candidate for the planed "marketplace" feature #811?

Ganesh-Chaudhari405:
yes,now it will only work for fedora but in future i will make it for all Linux distribution using ansible scripts.

PR #899, opened 2021-08-29
New_Plugin_ScreenShot

ShreyPatel4:
I am New to Open source
and I am stuck and need help with defining token/trigger words that activate my plugin
increment the progress from here

### DIFF
--- a/installer/requirements.txt
+++ b/installer/requirements.txt
@@ -44,3 +44,4 @@ Image
 img2pdf
 pdf2image
 python-nmap
+pyautogui
--- a/jarviscli/plugins/screenshot.py
+++ b/jarviscli/plugins/screenshot.py
@@ -1,0 +1,30 @@
+import pyautogui
+# some_file.py
+import sys
+# insert at 1, 0 is the script path (or '' in REPL)
+sys.path.insert(1, 'jarviscli')
+
+import CmdInterpreter
+import requests
+from colorama import Fore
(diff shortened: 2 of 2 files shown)


pnhofmann:
```
import pyautogui
import requests
from colorama import Fore
from plugin import plugin, require, alias
import os

@require(network=True)
@alias("ss", "capture")
@plugin("screenshot")
def ss(jarvis,s):
    jarvis.say("Please give me a moment", Fore.BLUE)
    incr += 1
    incr = str(incr) 
    myScreenshot = pyautogui.screenshot()
    myScreenshot.save(r'./../ss'+incr+'.png')
    if(os.path.isfile('./../ss'+incr+'.png')==False):
        jarvis.say("Success!", Fore.BLUE)
    else:
        jarvis.say("Sorry, an error occured retry using /ss , /screenshot, /capture", Fore.RED)
```

So, you don't have to call something like `` CmdInterpreter.input()`` - the plugin will be called if the user types screenshot and then the parameter s contains everything the user typed apart from screenshot.

Also no need to modify python path ;).

You would just need to initialize ``incr``. Probably use ``jarvis.get_data(key)`` and ``jarvis.update_data(key, data)``

https://github.com/sukeesh/Jarvis/blob/master/doc/API.md#get_data

ShreyPatel4:
yes, I will work on the error.

Issue #894, opened 2021-08-24
[Feature] telling Jarvis to Capture screen.


ShreyPatel4:
let's add this productivity feature to the AI agent also further we can use it to log it when issues or errors occurrence?

share ideas if any.

what I thought it add a simple rule on receiving input for multiple possible commands like "ss", "take ss", "screenshot", "capture screen ", etc it should perform a certain task for continued actions.

I have achieved the screenshot part. just don't know how to merge with existing files

```
/home/nan/Desktop/Open_source/Jarvis/env/lib/python3.8/site-packages/pydub/utils.py:170: RuntimeWarning: Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work
  warn("Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work", RuntimeWarning)
**No module named 'pyautogui'**
Jarvis' sound is by default disabled.
In order to let Jarvis talk out loud type: enable sound
Type 'help' for a list of available actions.
180 plugins loaded 12 plugins disabled. More information: status

~> Hi, what can I do for you?
screenshot
I could not identify your command...
~> What can I do for you?
```

pnhofmann:
Yeah, create PR and let us see. So, both @ShreyPatel4 and @26tanishabanik.

> I have achieved the screenshot part. just don't know how to merge with existing files

Looks like issue with virtualenv. ``**No module named 'pyautogui'**`` Jarvis can't find your module.

In jarvis home dir call ``source env/bin/activate``. Now you are in the Python-Environment Jarvis uses. Then you can install the package (again!) with pip like you are probably used to it (pip install pyautogui).

And put pyautogui in installer/requirements.txt ;).

ShreyPatel4:
@pnhofmann I have created PR and Guide me from there

26tanishabanik:
I have added a new feature in the dev branch. This is my Pull request https://github.com/sukeesh/Jarvis/pull/900

ShreyPatel4:
@26tanishabanik  that is great.
can you guide me where you added the token as trigger words for that plugin

26tanishabanik:
> @26tanishabanik that is great.
> can you guide me where you added the token as trigger words for that plugin

Yah sure, did you add the tag @plugin at the very beginning of the code after import to add the new plugin

ritikranjan12:
@ShreyPatel4 
I have made a plugin to capture the screen can i make a Pull Request?

[...]

PR #900, opened 2021-08-29
New feature Capture Screen

26tanishabanik:
Added new feature Capture Screen. Issue https://github.com/sukeesh/Jarvis/issues/894

### DIFF
--- a/jarviscli/plugins/screencapture.py
+++ b/jarviscli/plugins/screencapture.py
@@ -5,7 +5,7 @@ import subprocess
 import platform
 import pyautogui
 
-
+@require(native='scrot')
 @alias("capture", "screen")
 @plugin('screen_capture')
 def capture_screen(jarvis, s):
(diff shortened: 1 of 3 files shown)


ShreyPatel4:
this is great work.

suggestion:- if we first just made the textual version then implement for voice models.?

jarviscli/plugins/screencapture.py
```diff
@@ -0,0 +1,35 @@
...
+import platform
+import pyautogui
+
+
+@alias("capture", "screen")
+@plugin('screen_capture')
+def capture_screen(jarvis, s):
+
+    if (platform.system() == 'Linux'):
+        proc = subprocess.Popen(['dpkg',
```
pnhofmann, line 14: Hi!
good thing to handle these kind of exceptions. We should also make the installer aware of scrot / pyautogui.

But dpkg is only for ubuntu or debian. The way intended for these kind of checks is:

```
@require(native='scrot')
@plugin('screen_capture')
```

26tanishabanik:
> this is great work.
> 
> suggestion:- if we first just made the textual version then implement for voice models.?

@ShreyPatel4  I think it applies to both

pnhofmann:
Thank you!

Issue #905, opened 2021-10-18
Error with equations


siddharth123-tech:
Traceback (most recent call last):
  File "C:\Users\siddh\OneDrive\Documents\GitHub\Jarvis\jarviscli\CmdInterpreter.py", line 286, in try_do
    do(self, s)
  File "C:\Users\siddh\OneDrive\Documents\GitHub\Jarvis\jarviscli\plugin.py", line 208, in run
    self._backend[0](jarvis.get_api(), s)
  File "C:\Users\siddh\OneDrive\Documents\GitHub\Jarvis\jarviscli\plugins\evaluator.py", line 86, in equations
    calc(
  File "C:\Users\siddh\OneDrive\Documents\GitHub\Jarvis\jarviscli\plugins\evaluator.py", line 264, in calc
    result = result.evalf()
AttributeError: 'list' object has no attribute 'evalf'

![image](https://user-images.githubusercontent.com/82219365/137783751-c531a640-c884-46c8-b15c-19e85eea3f58.png)

pnhofmann:
Thanks for reporting this issue

viniciusdepadua:
Hi, is this issue still open? could I help? I'm new on contributing to open source projects. thx

pnhofmann:
#908

PR #908, opened 2021-11-04
fix: lambda expr error

viniciusdepadua:
Fixes issue #905. It was just a lambda function error. With this said, the function still can't solve linear equations


pnhofmann:
Thanks ;)

PR #914, opened 2021-11-20
Installer extra bash support

rijumone:
Will attempt to figure out the default shell of the current user. Will fallback to `bash` upon failure.

### DIFF
--- a/installer/helper.py
+++ b/installer/helper.py
@@ -172,3 +172,20 @@ def shell(cmd):
     sys.stdout.write(' \b')
 
     return exit_code
+
+
+def get_default_shell():
+    '''
+    Determine and return the default shell
+    of the current logged in user.
+    Args:
+        None
+    Returns:
+        shell name (str)
+    '''
+    try:
+        bin_path = os.environ.get('SHELL')
+        return bin_path.split(os.path.sep)[-1]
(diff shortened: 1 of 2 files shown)


pnhofmann:
First, nice implementation, I like it. Thank you!

Nevertheless, I think we should validate 'user_shell' against realistic values.

For example, the user might be using fish - which is using a (in my opinion) rather strange approach:
> set -Ua fish_user_paths /path
This code would break the users shell

Today the user might even be using stuff like powershell, which does not have something like .pwsh

In case the shell isn't recognized, the installer probably shouldn't suggest the option to modify .*.rc?

Milind220:
Sounds like a good idea, I think @pnhofmann is right though. The fallback option should be just to leave the .rc file alone

rijumone:
I concur.
Will update the implementation to detect current running shell and if it is not `bash` or `zsh,` not show the option to edit the `.rc` file.
Working towards detecting the `fish` (and other) shell now.
Thanks!

rijumone:
Pushed the changes, please review @pnhofmann, @Milind220.

pnhofmann:
All good, thank you!

rijumone:
> All good, thank you!

Awesome!

PR #930, opened 2022-02-20
Add hackernews plugin

coskundeniz:
This plugin lists titles from https://news.ycombinator.com/ and opens selected titles in the browser as separate tabs.


pnhofmann:
Nice.

Thank you!

PR #943, opened 2022-03-28
[New Feature] Image Compressor

### LINKED ISSUE #940: [New Feature] Image Compressor
The idea would be to @orlando-pt and I to implement an algorithm to compress images and have them occupy less space after processing.

The use-case will be for the user to ask Jarvis to use Image Compressor and then, input a path to a folder with images or a full image path to be compressed.

hugofpaiva:
Added a new feature w/ @Orlando-pt to compress images that closes #940 
Tests were also added where images are downloaded, compressed using Jarvis and the size of the images was verified to be less than the original.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ env/
 *.pyc
 *.log
 .idea/
+venv/
 MANIFEST.in
 setup.cfg
 setup.py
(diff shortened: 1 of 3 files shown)


pnhofmann:
Really nice work. Good code style, test and documentation - I like it.

Thank you!

Orlando-pt:
Thank you very much! We really like this project and there is a 100% chance that you will continue to see contributions from us :smile:

PR #947, opened 2022-04-02
[New Feature] Music Recognition

### LINKED ISSUE #945: [New Feature] Music Recognition
The idea would be to @orlando-pt and I to implement music recognition through Shazam's public API. For this, the [ShazamAPI](https://pypi.org/project/ShazamAPI/) or [ShazamIO](https://github.com/dotX12/ShazamIO) library will be used.

The use-case will be for the user to ask Jarvis to use Shazam and then, he hears the microphone sound (probably through [PyAudio](https://pypi.org/project/PyAudio/)) in the next 10-20 sec and sends the captured sound to the API to obtain a possible match.

hugofpaiva:
Added a new feature to do music recognition with the help of the available Shazam API that closes #945 

A test was also added where the microphone was mocked, a 15s excerpt of the Addams Family song is returned when the microphone is called and the response from Shazam is checked to be correct.

### DIFF
--- a/jarviscli/plugins/music_recognition.py
+++ b/jarviscli/plugins/music_recognition.py
@@ -1,0 +1,67 @@
+from io import BytesIO
+import tempfile
+from plugin import plugin, require, alias
+from colorama import Fore
+from shazamio import Shazam
+import pydub
+import asyncio
+
+requirements = ['ffmpeg']
+try:
+    import pyaudio
+    import speech_recognition as sr
+except ImportError:
+    requirements.append(
+        'voice_control_requirements (install portaudio + re-run setup.sh)')
+
+
+@require(network=True, native=requirements)
+@alias("shazam")
+@plugin("music recognition")
+class MusicRecognition:
+    """A tool to recognize music through Shazam API."""
+
+    def __init__(self):
+        self.sound_recorded = None
+        self.shazam = Shazam()
+
+    def __call__(self, jarvis, s):
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(self.music_recognition(jarvis))
+        
+
+    async def music_recognition(self, jarvis):
+        """Main function.
+
+        This function is the main menu, that will run
+        until the user says otherwise.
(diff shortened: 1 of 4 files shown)


hugofpaiva:
I want to ask for the opinion of the maintainers...

Using this song example:
- Song Title: Rock With You (Live)
- Song Artist: The Jacksons

The cover art is as following:
Cover art

However, the plugin I used to create a pixel art for the cover art draws the following:
Pixel art

Do you still want that I keep this drawing feature or is it too bad? I could increase the size of the drawing but will will occupy too much space 🤪

Orlando-pt (approved):
Nothing to point out. The implementation seems very good :fire: 

Lets wait for what the maintainers have to say about the quality of the image...

jarviscli/plugins/music_recognition.py
```diff
@@ -0,0 +1,210 @@
...
+import pydub
+import asyncio
+import climage
+
+requirements = ['ffmpeg']
+try:
+    import pyaudio
+    import speech_recognition as sr
+except ImportError:
+    requirements.append(
```
pnhofmann, line 17: Pretty clever, I like this idea. Yeah, this is better than anything I thought of; I'll use this pattern for other plugins, too ;).

pnhofmann:
> Do you still want that I keep this drawing feature or is it too bad?

I like it. That's just how images on the cli look like.
I'm thinking, maybe I'll add a 'print_image()' method to the jarvis "core" at some point, which might be able to (optionally) use feh or something. But at this point, the image is all right,.

 So.

Hey, your plugin is amazing! I like it a lot. This is a feature, which makes so much sense for a "personal assistant" and it works so good.

So, thank you ;).

(And bye the way sorry for the 16 days delay - how time flies...)

hugofpaiva:
Thank you for your feedback and for the compliments! This project is really great to work on because of how well thought the code structure is set in place.

Don't worry about the delay, our life is not just coding 🤪

PR #951, opened 2022-04-12
Fix flake8 erros when running ./test.sh

AirlineDog:
I noticed that when i run the `test.sh` file a large number of flake8 errors apear.

I changed the formatting of the specified files so there are no such erros any more.


pnhofmann:
Thank you

Issue #944, opened 2022-03-29
Improve calories plugin


Papajohn77:
I suggest the following improvements for the calories plugin:
- Not suggesting calorie intake less than 1200 cal for females and 1500 cal for males (Harvard Health Publications suggests women get at least 1,200 calories and men get at least 1,500 calories a day unless supervised by doctors). Instead we can suggest them to increase their activity level.
- Suggesting the appropriate macronutrients (proteins, carbs & fats) for losing or gaining weight.
- Providing some more information to the user.

I would like this to be assigned to me if you agree with the above suggestions.

pnhofmann:
sure, go for it!

PR #952, opened 2022-04-18
Improve calories plugin #944

Papajohn77:
I've implemented the suggested improvements mentioned in #944

The CaloriesMacrosPlugin class is now responsible to read and validate the user input & start the plugin. I created the CalorieCalculator inner class that is responsible to calculate the recommended daily calorie intake based on the data provided by the user & also the MacronutrientCalculator inner class that is responsible to calculate the macronutrients that correspond to the calorie intake of the user. Furthermore, i have created unit tests for all of the methods.


pnhofmann:
Nice. Looks good, thank you!

PR #957, opened 2022-04-22
created beer.py

JRenard97:
This is a plugin that sends a user a random beer

### DIFF
--- a/jarviscli/plugins/beer.py
+++ b/jarviscli/plugins/beer.py
@@ -1,0 +1,30 @@
+import os
+import requests
+from colorama import Fore
+
+# Generate a random beer for the user 
+@require(network=True)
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/beer.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # Pull from a random beer API
+    rx = requests.get('https://api.punkapi.com/v2/beers/random')
+    x = 0
+
+    # Gross JSON fixing, needs to be refactored
+    for i in rx:
+        f = f + str(i)
+        x = x + 1
+        if x == 2
```
pnhofmann, line 22: This is definitly a SyntaxError and I don't really understand what you are trying to do here...

PR #959, opened 2022-04-24
Update Weather Report Plugin

dustinn1:
# Update Weather Report Plugin

##  Description

An update to the weather report plugin using the OpenWeatherMap API to display more information and improve functionality.

##  Updates
- Switch to using OpenWeatherMap API
- Display a list of cities that matches or are similar to the user's query
- Give the option of using either imperial (°F & mph) or metric (°C & km/h) units
- Display weather data for the current day and the next 7 days 

## Screenshots
List of cities that matches or are similar to the user's query
List of cities that matches query

Example of looking up the forecast for a city (only first 2 days are shown)
Weather forecast lookup example

##  Contributors
- @Dev-Will-R
- @asteraceae
- @dustinn1
- @jgo1LUC

### DIFF
--- a/jarviscli/plugins/weather_report.py
+++ b/jarviscli/plugins/weather_report.py
@@ -1,51 +1,91 @@
+#Import necessary packages
 import requests
 from plugin import plugin, alias, require
 from colorama import Fore
-
+import datetime
 
 @require(network=True)
 @alias("forecast")
 @plugin("weather report")
+
 class WeatherReport:
-    """The user will input a location as a string, and the WeatherDB database will be used to make a GET request and
-    fetch data. Then, the user can see the weather forecast for the upcoming week if they wish. """
+    """Get the weather forecast for the current day and next 7 days. The user can input a location and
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/weather_report.py
```diff
@@ -1,51 +1,91 @@
...
-    """The user will input a location as a string, and the WeatherDB database will be used to make a GET request and
-    fetch data. Then, the user can see the weather forecast for the upcoming week if they wish. """
+    """Get the weather forecast for the current day and next 7 days. The user can input a location and
+        recieve weather data from OpenWeatherMap API."""
 
-    def __call__(self, jarvis: "JarvisAPI", s: str) -> None:
-        self.print_weather(jarvis)
+    # Fetch information using API key
+    def __call__(self, jarvis, s: str) -> None:
+        self.apikey = "c15e25198031e54d78ef4740a7b53e5f"
```
pnhofmann, line 17: Hmm, your api key?
Not the best way to embed API keys in source code

PR #963, opened 2022-05-05
[New Feature] Balut dice game #955

### LINKED ISSUE #955: [New Feature] Balut dice game
If that's ok i would like to create a plugin for playing the [Balut](htts://www.balut.org/rules.html) dice game.

Papajohn77:
The Balut plugin is finally ready! This pull request closes #955. 

In this plugin i chose to use the python's built-in methods input & print, because the JarvisAPI methods input & say have some limitations. For example for printing the results i couldn't pass the end argument to the say method and the alternative solution in order to to display the results the way I wanted was unnecessarily complicating the code.

### DIFF
--- a/jarviscli/plugins/balut.py
+++ b/jarviscli/plugins/balut.py
@@ -1,3 +1,6 @@
+import random
+import textwrap
+import itertools
 import functools
 from plugin import plugin
 from typing import Callable, List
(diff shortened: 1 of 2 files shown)


pnhofmann:
Nice one! With tests and typehints and everything. Good work.
Thanks!

PR #969, opened 2022-05-24
[New Feature] QR generator for URLs

### LINKED ISSUE #962: [New Feature] QR generator for URLs
Hello, i had this idea for a plugin that creates a QR code for a URL that the user gives as input. I will use the following [api](https://goqr.me/api/) . This api creates a PNG image of the QR ,that Jarvis will save on the directory the user specifies.

george-kakavas:
# QR Generator

I've added this new feature that creates a QR code for a URL the users specifies.The QR code is stored as a png image. This feature closes #962

## Steps
- User gives a **valid (existing) URL**

qr3(1)

- User gives a **correct filepath (directory)** for the QR to be stored

qr3

- User gives the **filename** and the execution is completed

qr3(2)

## Result
A png image containing a **QR code** that leads to **Jarvi's central page on Github** is stored in the specified directory with the name of the user's choice. 

## Added Functionality
- The URL the user gives, is checked for it's reachability. If the URL can't be reached (**a number of exceptions is covered in the plugin,for example typos in the url**) then the user is asked to type it again or **type "end" to exit** as seen bellow. 

qr1

- The filepath where the QR will be stored is checked using **os.path.isdir()** . If the directory doesn't exist, the user is asked to enter it again.

qr2

- The filename the user gives is **properly named (whitespaces are replaced with underscores)**

### DIFF
--- a/jarviscli/plugins/qr_generator.py
+++ b/jarviscli/plugins/qr_generator.py
@@ -1,0 +1,168 @@
+
+from plugin import plugin,require
+from colorama import Fore
+import requests
+import os
+import re
+
+
+@require(network=True)
+@plugin("qr")
+
+class QRGenerator:
+    """
+    Provides you with a png image of a QR for a specified URL
+
+    Enter 'qr' or 'QR' to use:
+    
+
+    Attribution:
+        QR generated from https://goqr.me/api/
+    """
+    
+    def __init__(self):
+        self.url = "http://api.qrserver.com/v1/create-qr-code/"
+    
+
+    def __call__(self,jarvis,s):
+        if s == 'help':
+            self._print_help(jarvis)
+            return
+
+        jarvis.say("\ntype 'qr  help' for additional information")
+        jarvis.say("")
+
+        prompt_url = ("\nType the URL that you want to create a QR code \n" 
+                      "or 'end' if you wish to exit: ")
+        # get url from user prompt
+        url = self._get_url(prompt_url,jarvis)
+        if url == "end":
+            return
+        jarvis.say("")
+
+        prompt_path = ("\nType the filepath where you want the QR to be stored \n"
+                       "or 'end' if you wish to exit: ")
+        # get path from user prompt
+        self.path = self._get_path(prompt_path,jarvis)
+        if self.path == "end":
+            return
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/qr_generator.py
```diff
@@ -0,0 +1,168 @@
...
+            return
+        jarvis.say("")
+
+        prompt_path = ("\nType the filepath where you want the QR to be stored \n"
+                       "or 'end' if you wish to exit: ")
+        # get path from user prompt
+        self.path = self._get_path(prompt_path,jarvis)
+        if self.path == "end":
+            return
+        self.path = self.path + "\\"
```
pnhofmann, line 49: Does not work on Linux.
Better use:

os.path.join(self.path, self.filename + '.png') 

in line 130

jarviscli/plugins/qr_generator.py
```diff
@@ -0,0 +1,168 @@
...
+            # send request
+            jarvis.spinner_start('Creating QR ')
+            response = requests.get(query)
+            # where the png file will be stored
+            location = self.path + self.filename + ".png"
+            file = open(location, "wb")
+            file.write(response.content)
+            jarvis.spinner_stop()
+            file.close
+        except BaseException:
```
pnhofmann, line 135: try-excepting any Exception shadows what kind of exception accrued and makes stuff harder to debug. For example I had Exception:
``[Errno 13] Permission denied: '/tmp\\x.png'``

which was definitely no API error ;). I would recomend:
* Min ``except BaseException as e`` and then ``print(e)``
* Better: Check what exception can actually be thrown by api errors and only catch those

pnhofmann:
Has some weird formation (trailing whitespaces; maybe check ``flake8 jarviscli/plugins/qr_generator.py``

But apart from that looks good!

pnhofmann:
Thanks

george-kakavas:
Great... I  actually wanted to fix the other issues you specified,before merging. 
The **commit**  f017111aadaed7647d941f41912e3f2cdb2e1576 was just for the first issue. I fixed only that in the little time i had and i intended to fix the other two in the following days. If you are still interested, i would be glad to make the remaining corrections.

PR #972, opened 2022-05-29
skip manual installation (virtualenv)

eldinesh:
If virtualenv module is not present in user's machine, it'll skip manual installation message, and perform installation.

### DIFF
--- a/installer/helper.py
+++ b/installer/helper.py
@@ -3,6 +3,7 @@ import sys
 import distutils.spawn
 import time
 import subprocess
+from sys import platform
 from threading import Thread
 from tempfile import NamedTemporaryFile
 from unix_windows import IS_WIN
@@ -18,6 +19,13 @@ try:
 except NameError:
     pass
 
+def install(package):
+    '''Install a package using pip or apt depending upon the platform'''
+    if platform == "linux" or platform == "linux2":
+        subprocess.check_call([sys.executable, "-m", "apt", "-q", "install", package])
+    elif platform == "darwin" or platform == "win32":
+        subprocess.check_call([sys.executable, "-m", "pip", "-q", "install", package])
+    return True
 
 def executable_exists(name):
     binary_path = distutils.spawn.find_executable(name)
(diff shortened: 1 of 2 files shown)


installer/steps/a_setup_virtualenv.py
```diff
@@ -22,12 +22,11 @@
...
-        printlog("Please install virtualenv!")
-        printlog("https://github.com/pypa/virtualenv")
-        printlog(unix_windows.VIRTUALENV_INSTALL_MSG)
-        printlog("")
-
-    fail(">>> Apparently virtualenv on Python3 ({}) does not work. Exiting!".format(unix_windows.VIRTUALENV_CMD))
+        printlog("virtualenv not installed, installing...")
+        if install("virtualenv"):
+            printlog("Installation Complete!")
+        else:
```
pnhofmann, line 28: Since install can't return anything else except True, the else-block cannot be executed

installer/helper.py
```diff
@@ -18,6 +19,13 @@
 except NameError:
     pass
 
+def install(package):
+    '''Install a package using pip or apt depending upon the platform'''
+    if platform == "linux" or platform == "linux2":
+        subprocess.check_call([sys.executable, "-m", "apt", "-q", "install", package])
```
pnhofmann, line 25: Hmm, apt is only for Debian/Ubuntu based distros. Won't work on Fedora, openSUSE, Arch, Gentoo, ...

eldinesh: We can use **DISTRO** package to get distro id, and then use package manager acc. to distor id. (https://distro.readthedocs.io/en/latest/)

OR

We can add an statement for other users to use package manager based on their distros manually.

OR

We can make these options by default, and get user choice based on their distro.

installer/steps/a_setup_virtualenv.py
eldinesh: okay, will rewrite it.

pnhofmann:
Think this gets to complicated; the user just have to install virtualenv by themself.

PR #975, opened 2022-06-06
Add Command List PDF

smitcha4:
Current Jarvis command list file for potential contributors and users.


pnhofmann:
Hi!
how do you make this PDF? Just having a PDF makes it hard (actually impossible) to maintain...

PR #981, opened 2022-06-17
Create word game

MinSeoke-Kim:
Five characters word guessing game.

### DIFF
--- a/jarviscli/plugins/wordgame.py
+++ b/jarviscli/plugins/wordgame.py
@@ -1,0 +1,111 @@
+from plugin import plugin
+import random
+from tkinter import *
+from tkinter import messagebox 
+
+@plugin("wordgame")
+with open("./Jarvis/jarviscli/data/wordgame_word_list.txt", "r") as wordslist:
+    data = wordslist.read()
+    words = data.split()
+    word_pos = random.randint(0, len(words) - 1)
+    random_word = words[word_pos]
+
+root = Tk()
+
+GREEN = "#039108"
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/wordgame.py
```diff
@@ -0,0 +1,111 @@
...
+word_guessed = []
+
+characters_left = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k',
+                   'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
+                   'w', 'x', 'y', 'z']
+
+
+def get_guess():
+
+    global random_word
```
pnhofmann, line 40: You don't need to 'global' variables if you don' write them. In fact, you should avoid writing global variables.

jarviscli/plugins/wordgame.py
```diff
@@ -0,0 +1,111 @@
+from plugin import plugin
+import random
+from tkinter import *
+from tkinter import messagebox 
+
+@plugin("wordgame")
```
pnhofmann, line 6: This is not really a plugin - which must be a function or executable class

jarviscli/plugins/wordgame.py
```diff
@@ -0,0 +1,111 @@
+from plugin import plugin
+import random
+from tkinter import *
```
pnhofmann, line 3: Hmm, Jarvis is actually CLI-based...

PR #984, opened 2022-06-19
Resolves #983

### LINKED ISSUE #983: Twitter Trends Plugin is down.
The twitter trends plugin is down because the api it uses is also down. We are in contact with the api maintainer to solve the issue. Except a fix soon from me and @LeonidhaMara.

LeonidhaMara:
Fix twitter trends plugin by updating the API.


pnhofmann:
Thank you!

PR #989, opened 2022-07-05
Prompt Generator Add.

Lallybroch:
added a prompt generator to Javis and made it editable through 'prompts.json' in the data folder.


pnhofmann:
Like it ;) Thanks

PR #1013, opened 2022-12-06
new horoscope plugin

MaelCossec:
Hello, I just created a plugin that gives someone's horscope. Thank you !

### DIFF
--- a/jarviscli/plugins/horoscope.py
+++ b/jarviscli/plugins/horoscope.py
@@ -1,0 +1,78 @@
+from plugin import plugin, require, alias
+import pyaztro
+
+@require(network=True)
+@alias("astro")
+@plugin("horoscope")
+class Horoscope :
+
+    def __call__(self, jarvis: "JarvisAPI", s: str) -> None:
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/horoscope.py
```diff
@@ -0,0 +1,78 @@
...
+        jarvis.say("11: Aquarius")
+        jarvis.say("12: Pisces")
+        sign = int(jarvis.input("Enter your number: "))
+        if(sign<1 or sign>12) :
+            jarvis.say("Wrong number input (Should be between 1 and 12). ")
+        else :
+            self.info_sign(jarvis, sign)
+
+    def info_sign(self, jarvis: "JarvisAPI",n : int) :
+        match n :
```
pnhofmann, line 33: Oh, match is new in python 3.10. Which is a nice feature and read great - but for backward-compatibility, a classic if-elsif-else would be better

pnhofmann (commented):
Looks good though, thanks!

PR #1016, opened 2022-12-08
Add the plugin music to listen music from Youtube

AlexisGrandjacquot:
To listen music from Youtube  on Jarvis. There is also a system to manage PlayList


pnhofmann:
Hey, thanks.
Your code looks good. Just have some problems. I'm just not so happy about the requirements.
* Don't want to mess with youtube toc
* pygame don't work in cli and needs to be installed
* also vlc does require vlc installed and compiled ctypes code

PR #1020, opened 2022-12-19
Added Random Fact Feature

vrrusso:


pnhofmann:
I was so free removing the unrelated commits

Thanks! Fun Fact: I actually really like random facts ;)

Issue #1021, opened 2022-12-21
[New Feature] What day of the week is any day


PauloMRocha:
Add a plugin to calculate what day of the week is any day in past or future.

pnhofmann:
merged #1022

PR #1022, opened 2022-12-21
Add plugin to calc. weekday of any date

PauloMRocha:
Add a plugin that calculate the day of the week for any day in past or future. #1021


pnhofmann:
That is pretty usefull! Thank you!

Issue #1010, opened 2022-11-29
Suggestion for Goodreads feature


ApostolidouK:
I am a bookworm that uses Goodreads (https://www.goodreads.com/) all the time and it is a great and popular toll for searching book reviews, new titles, authors, book suggestions, etc. So, can I work on adding a plugin for it to this project?
My idea is that the user could search this way for book reviews, authors, suggestions per genre, the most popular books every day/month (and any other ideas are welcomed).

pnhofmann:
Hey, hmm for services like this it's always the trouble with api keys. And scrapping probably violets TOC and I don't want to mess with amazon.

But basically **yes**, no problem. Would also be okay to ask the user to provide own api key.

ApostolidouK:
I looked into it a bit more and it looks like the website stopped issueing new api keys. So, instead I am thinking that the actions (most read this month, search, suggestions, etc) will be suggested to the user and then once they choose what they want to see and the parameters (like title, search by author or by genre, newly published, etc) if necessary, they will be directed to the corresponding Goodreads page. It can be a tricky page to figure out, especially for a new user because some features are a bit disorganized, but this way it will be more accessible since you will be immediately directed to the thing you are looking for.

> Hey, hmm for services like this it's always the trouble with api keys. And scrapping probably violets TOC and I don't want to mess with amazon.
> 
> But basically **yes**, no problem. Would also be okay to ask the user to provide own api key.

pnhofmann:
> if necessary, they will be directed to the corresponding Goodreads page

That would work nicely

ApostolidouK:
I finished the feature, I tested all the parts to make sure everything is running smoothly with the rest of the code and I made a new pull request.


> > if necessary, they will be directed to the corresponding Goodreads page
> 
> That would work nicely

pnhofmann:
merged #1024

PR #1024, opened 2022-12-21
Goodreads Feature. Issue #1010

ApostolidouK:
Goodreads Feature. Issue #1010
The user is able to check for the most popular books each week, search a book or author, look for suggestions by genre and even have a look at running giveaways from authors or publishers.


pnhofmann:
That's nice, thanks

Issue #693, opened 2020-05-17
plugin String "s" is always lowercase (MACOS)


evamy:
found out trying to add a whoami plugin for macos.
the string being passed (s) into the function is always lowercase, no matter what options I pass

madhavkulkarni1986:
@evamy Yes, because all the input data is converted to lower case in jarviscli/Jarvis.py
line: 125         data = data.lower()

I do not know why was it designed that way. May be @pnhofmann or @sukeesh or @appi147 can answer. But almost of the plugins comply with this now. So, removing the translation to lower case will require all the plugins using this feature to be changed.

However, this is debatable. Please put forward your case that requires input's case to be same as it is passed to the function/plugin.

evamy:
@madhavkulkarni1986 thanks for that info.

What happens to commands that can have uppercase options ? Those will never be taken into consideration at all. Specially like "id" command on MACOS; it takes only single letter command arguments, most of which are uppercase.

```
id: illegal option -- f
usage: id [user]
       id -A
       id -F [user]
       id -G [-n] [user]
       id -M
       id -P [user]
       id -g [-nr] [user]
       id -p [user]
       id -u [-nr] [user]
```

pnhofmann:
You are absolutely right!
Similar issue with stuff like 'dot' - which are also removed. Which is good for most plugins, but there are some exceptions.

We should probably add some kind of "flags" to change this behavior. Maybe something like:

```
@feature(CASE_SENSITIVE, PUNCTUATION)
@plugin("helloworld")
def helloworld(jarvis, s):
    """Repeats what you type (case sensitive)"""
    jarvis.say(s)
```

Any suggestions?

evamy:
@pnhofmann 
Flags actually sounds a good idea.
I have not used that before in python, but let me look and try it out. See if it works, while also trying to understand how much change it would take to do this.

madhavkulkarni1986:
@pnhofmann I added the necessary class and function to implement this feature. But somehow I am not able to get the flags, via the plugin class. In fact, I am not able to get the plugin under question.
My commit is here: https://github.com/madhavkulkarni1986/Jarvis/tree/feature_lower_case_input

Can you please check and let me know how do I get the feature of one particular plugin? Or can you point out how to fetch the called plugin class? Say I asked for "quote" on jarvis prompt, how do I get quote class?

evamy:
[...]

PR #1027, opened 2023-01-02
Mnist classifier

markioan01:


pnhofmann:
Hi!

---------

yeah - the ``JarvisMnistClassifier.py`` file must be in ``jarviscli/plugins`` to be actually loaded.
Also ``MyMnistModel.pth`` should be placed in ``jarviscli/data``.

---------

Please also note, that torch, torchvision and maybe also opencv and mathplotlib must be installed in Jarvis virtualenv (systemwide installation only won't work).

```
source env/bin/activate
pip3 install torch torchvision
pip3 install ...
```

which also means we'll need to modify the installer-script - but that won't be hard.

---------

Yeah, please notice #693 - the String s is always lowercase. I'll fix this bug - eventually, so don't worry. But as a workaround - for now - use lower case only filenames.

Independent from this issue, you should check if the filepath given by the user actually exists and if not print a more meaningful error message.

---------

Next, you are using:

```
try:
    # (...)
except:
    jarvis.say("somthing went wront")
```

You shouldn't do this because it makes debugging way harder. Error messages are important tools to understand what actually went wrong and by using this pattern you are "shadowing" useful output. In this case:

```Some error occurred, please open an issue on github!
Here is error:

  File "/data/git/Jarvis/jarviscli/CmdInterpreter.py", line 286, in try_do
    do(self, s)
  File "/data/git/Jarvis/jarviscli/plugin.py", line 208, in run
  File "/data/git/Jarvis/jarviscli/plugins/mnist_classifier.py", line 50, in mnist_classifier
    if(not(img[row][col][0] == pix[row][col][1] and pix[row][col][1] == pix[row][col][2])):
UnboundLocalError: local variable 'pix' referenced before assignment
````

---------

That all being said, your stuff certainly works! Yep, it's nice.

**Comment by markioan01** [comment]
Thank you very much for your help! As I said it was my first contribution to a public git repo and you were really helpfull.
I should try to fix these small problems or you have already handled them?

**Comment by pnhofmann** [comment]
> Should I try to fix these small problems or you have already handled them?

Okay, I can just push what I have. Also fixed (?) the black/white check that was broken. Yeah, what is actually left for you to do?

* Documentation, like maybe add a description
```
@plugin('classify')
"""
Your description
"""

* Print usefull information if user inputs invalid picture path

* Maybe scale down image to 28x28 size if user inputs image with different sizes?


------

What I will do:

* Update installation script

markioan01:
I made the changes!

markioan01:
Maybe If we put the @plugin("classify") above the myNN class so it reads it before running the def

pnhofmann:
Thanks

PR #1033, opened 2023-01-21
Adding the option to play tic_tac_toe game against Jarvis.

zaidisyed1991:
Adding the option to play tic_tac_toe game against Jarvis. Jarvis will be using the minmax algorithm to play the best possible moves. Jarvis will take the position of the 'O' player (the second player).


pnhofmann:
Nice.

Works good - maybe even a bit too good, since you can't really win, can you?

Basically, thank you; I nothing to add and will merge now.

PR #1035, opened 2023-01-29
Add a new feature

ampatzidhs:
Adding guess the number feature


pnhofmann:
Nice visualization!
I like it a lot. That's actually perfect for teaching how binary search works ;).

Thanks a lot!

------
------

Only one minor hint: "Add guess the number feature" would be the better headline for the PR.

PR #1039, opened 2023-02-08
kaprekar plugin

mkatsara:
I created a simple, new plugin for Jarvis. You can access the command from the terminal window with the word 'kaprekar'. Give a number and the algorithm checks if it is Kaprekar number or not.


pnhofmann:
Thanks!

PR #1042, opened 2023-02-08
created a christmas timer (new upload)

pap35000:


jarviscli/christmas timer.py
```diff
@@ -0,0 +1,47 @@
+from plugin import plugin, pap35000
```
pnhofmann, line 1: Why are you importing ```pap35000``` and what is this supposed to do?

Also, file needs to be placed in jarviscli/pluings to be loaded.

pap35000: how I place it to jarviscli/pluings

PR #1043, opened 2023-02-08
Add the missing requirements

AbdoullahBougataya:
Add the missing requirements:
_PyPDF2
_pyautogui
_SpeechRecognition
_fasttext


pnhofmann:
Hey thanks, but these requirements are actually installed here:
https://github.com/sukeesh/Jarvis/blob/master/installer/optional.py

They all require native non-python extension which needs some extra handling - otherwise this makes problems on some systems.

PR #1044, opened 2023-02-09
new periodic elements plugin

argyro2:
Type the name of an element of the periodic table to get back its symbol, atomic number, atomic mass and group.


pnhofmann:
That's nice; thank you!

PR #1046, opened 2023-02-11
Feature/plugin distance calculator

FlynnOwen:
I've added a plugin to calculate the haversine distance between two countries supplied. Uses the rest countries API: https://restcountries.com/

Let me know if there are any changes I could make, or whether tests (unit or other) should be added.


pnhofmann:
I like it!
Thanks ;).

PR #1061, opened 2023-03-16
Fixed typo on history call

sabari50312:
Changed  
def histroy_call(self):  
to
def history_call(self):


pnhofmann:
Thanks, but already merged in #1056

Issue #1065, opened 2023-03-29
Typing-test feature not working


malhaar2002:
The "typingtest" feature is resulting in an error. I would like to work on this. Thank you.

![image](https://user-images.githubusercontent.com/58373047/228525643-f42f0e79-caf9-4cf6-b2c8-0c95dc967008.png)

pnhofmann:
#1066

PR #1066, opened 2023-03-29
Fix Bug Regarding Jarvis Typingtest

malhaar2002:
I have made the following changes:

1. Fixed issue #1065 by changing the random-words API being used. It seems like the previous API is no longer supported.
2. Changed the number of words required for the typing test from 4 to 10, as 4 words are not sufficient for conclusive results
3. Removed unused imports


pnhofmann:
Thanks!

PR #1077, opened 2023-04-11
Modify Spinwheel so that it has a spinwheel GUI instead of just showing output in terminal

spacecowbye:
This was the output before
spinwheelbefore

I added a create spinwheel function which will display the output in a proper spinwheel image format
spinwheelafter

### DIFF
--- a/jarviscli/plugins/spinthewheel.py
+++ b/jarviscli/plugins/spinthewheel.py
@@ -6,7 +6,7 @@ import math
 
 def create_wheel(sectors):
     """
-    Creates GUI of a spinwheel, previous code was simply used to show the output in the terminal.
+    Creates GUI of a spinwheel and closes the window after 5 seconds.
 
     """
     # Define the size of the canvas and the radius of the wheel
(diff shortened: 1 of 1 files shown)


pnhofmann:
Hey, year - works good. I do like it, my only concern is, it (obviously) doesn't work if you are working CLI-only without GUI. 

I'll probably add a ``require(XSERVER)`` ore something. Comment here, should I forget ;).

Additionally, a close function is required, otherwise the window will stay open forever + block further Jarvis operations
https://stackoverflow.com/questions/8009176/function-to-close-the-window-in-tkinter

spacecowbye:
yeah sure thanks!! I dont know what to do if I am  are working CLI-only without GUI, I can add the close function right now!!

spacecowbye:
Added the line to destroy the window after 3 seconds!!

spacecowbye:
> Hey, year - works good. I do like it, my only concern is, it (obviously) doesn't work if you are working CLI-only without GUI.
> 
> I'll probably add a `require(XSERVER)` ore something. Comment here, should I forget ;).
> 
> Additionally, a close function is required, otherwise the window will stay open forever + block further Jarvis operations https://stackoverflow.com/questions/8009176/function-to-close-the-window-in-tkinter

hello can you please close the issue and merge the pr

Issue #1069, opened 2023-04-03
Rename reboot function to 'reboot computer'


dinanz:
Currently the reboot function only reads 'reboot' and if someone uses it unaware that it reboots the entire computer and not Jarvis, they may lose unsaved data. I'd like to work on renaming this feature to be something more explicit that describes that it reboots entire computer system.

pnhofmann:
Sure!

PrettyAX:
I'm gonna submit a pull request that adds some additional measures regarding asking for user confirmation regarding rebooting the system

PR #1083, opened 2023-04-18
Update shutdown.py

PrettyAX:
Corrected the name from reboot to reboot_computer. Added additional functionality where it prompts the user to confirm reboot process.

This addresses issue #1069


pnhofmann:
makes sence

Issue #1068, opened 2023-04-03
Change font colour based on terminal background


dinanz:
Suggestion to add feature to change text colour based on terminal background colour or system theme, if possible. Currently when the background is black, the text that are in dark blue and red are almost unreadable. I'd like to suggest and try working on this: if the terminal background is detected to be black or system is set to dark theme, changee the colours to be much brighter so they are easier to read.
![image](https://user-images.githubusercontent.com/60189142/229392481-1eb5f128-ccec-4154-bd56-0f5356c5cae5.png)

pnhofmann:
Yep, that would be nice. 

I actually don't really know if detecting background color of Terminals works corss-platform (and in all kind of terminal-emulators?).
If not, manual configuration of "light" and "dark" mode would also work fine for me.

Pinned this issue, since I agree we actually need this feature. I'm happy for any input ;).

Somehow related: 485

dinanz:
I looked at #485 and I can see that it's a little complex to configure personalisations especially with storage... I can definitely look into detecting OS and then different functions for detecting theme based on OS if that's possible at all, and if not I suppose I can try (at least to begin with) perhaps a temporary configuration for the text colour.
Thanks! :)

pnhofmann:
That's actually a good compromise

JustinMLu:
I'm gonna try to make a plugin that allows Jarvis to accept text commands to change text color (or just automatically does it), if that's okay with you guys? Looking to be done by either tomorrow or the day after

dinanz:
@JustinMLu Hi, I'm in the middle of changing the default text colors so I'm going to finish this, but it's only minor fixes in the main file and in pluginManager. But afterwards I think if you can work on a manual config that would be great, and as a separate plugin I don't think it's going to conflict with my changes either. Thanks!

JustinMLu:
No worries! I'm actually nearly finished with my plugin that lets you change colors manually & I don't wanna step on your toes.

I don't think our changes will conflict unless you're changing the default text colors by hard-coding the colorama constants (my default text colors are stored in a JSON so they can be changed).

 My plugin also adds support for preset themes (and resetting to default), so I could even add your chosen theme as the 'default theme' afterwards if you wanted :)

JustinMLu:
Update: I might make a more formal issue later, but I've noticed some issues with the "_first_reaction_" boolean (in Jarvis.py and CmdInterpreter.py).

[...]

PR #1084, opened 2023-04-19
Change default font colors to be suitable for both white and black terminal background

dinanz:
Attempts to resolve issue #1068


pnhofmann:
Thanks!

PR #1087, opened 2023-04-21
Theme update & plugin! (Preset & custom persistent) themes)

JustinMLu:
By editing some colorama properties located in jarviscli and building my own plugin, I've managed to add functionality onto Jarvis that allows for **persistent user themes! WIth this theme, you can:**

- Manually created using an interactive theme-editor plugin
- Saved persistently and changed immediately after user specification
- Picked from a list of custom-made preset themes (including lightmode & darkmode)

Currently, these themes are only functional for non-"custom" plugin interactions, given the nature of colors being set via colorama & the sheer amount of plugins present.

Please let me know what you think and if you have any ideas/concerns!


JustinMLu:
**Update:** Due to some manual config-ing I did for my University's project report, I accidentally broke the plugin's ability to store json files (it was working up to a day ago). Please don't reject this yet, I'll fix it immediately once the report is written tonight

pnhofmann:
Oh, this is nice! Good work! I especially like the pre-defined themes.
Thanks a lot.
Just write me, when it's ready to merge. And please write again, should I forget / miss ;).

JustinMLu:
Hey @pnhofmann , thanks for the positive feedback! The issue I'm struggling with right now is specifying the path to the JSON theme file ("user_theme.json") for both GeneralUtilities.py and in the actual change_color.py plugin script; relative path-ing works fine in my Python 3.10.10 venv but when I try it in the actual Terminal interface I encounter a file-not-found error.

PR #1090, opened 2023-04-25
New Plugin Chat

jhutchings99:
This pull request adds a new feature to the project: a plugin that uses GPT-2 and gpt-2-simple to generate responses based on user input. The plugin also allows users to choose which GPT-2 model they want to download before using it.

Added a new plugin that uses GPT-2 and gpt-2-simple to generate responses based on user input.
Implemented the ability for users to choose which GPT-2 model they want to download before using it.


pnhofmann:
I like it. Thanks!

But how I see, it's more a text generation tool - which is obviously fine, since gpt is a text generator. Maybe naming it "tell story" would be more fitting?

Also, I'll need to make view adjustments to installer, since even installing python package ``gpt-2-simple`` with requirements is quite big and if a user doesn't want to use this feature anyway, I want to give them the option to skip installation.

Yeah, but thanks for actually implementing this!

PR #1094, opened 2023-04-29
Added jumble solver plugin

gaganchandan:
Added a new plugin called `jumble` which can be used to unscramble words. Source can be found in `jumble.py` in the `plugins` directory. No new dependencies are required.


pnhofmann:
Hey, thanks! I like it.

Just small question, could you add a small comment about the words-file - like where this file is from and how it was generated. Basically both curiosity and maybe this information could be used to update the word-file someday?

PR #1097, opened 2023-05-01
add new plugin for string conversion

dinanz:
Added a new simple plugin/feature that allows an input string and converts it to upper/lower/title/sentence case.


dinanz:
Closes issue #1100

pnhofmann:
thanks!

Issue #1102, opened 2023-05-02
The cryptotracker feature doesn't work


PrettyAX:
Cryptonator is defunct and no longer works, thus the feature is now defunct

PrettyAX:
Push an update to use a different API to achieve the same results

PR #1103, opened 2023-05-02
Update to Cryptotracker

PrettyAX:
Implemented a new API from API-ninjas that is able to return the favorite pairs of cryptocurrency and their USD prices. If the user inputs a pair that is not supported, it will query a different API to print out all of the pairs that are supported.

Addresses issue #1102

### DIFF
--- a/jarviscli/plugins/cryptotracker.py
+++ b/jarviscli/plugins/cryptotracker.py
@@ -13,27 +13,13 @@ FAVORITES = [
 ]
 
 
-def print_in_color(change):
-    """
-    Prints the price change with color. If the price is down, print in red.
-    If the price is up, print in green.
 
-    Parameters
-    ----------
-    change: str
-        The price change.
-    """
-    if float(change) < 0:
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/cryptotracker.py
```diff
@@ -44,38 +30,44 @@ def check_prices(base, target):
...
+    url = 'https://api.api-ninjas.com/v1/cryptoprice?symbol={}{}'.format(
         base.lower(), target.lower())
 
     try:
         response = requests.get(
-            url, headers={
-                'User-Agent': 'Jarvis'}).json()
-        price = response['ticker']['price']
-        change = response['ticker']['change']
+            url, headers={'X-Api-Key': 'NtgQxUQ02CDzcA5It1e/0w==25CTvSFoQ2nBBCSm'}).json()
```
pnhofmann, line 38: This api key, is it yours? Hmm, not so nice.

Issue #1109, opened 2023-05-03
Password Generator


Pasindu3683:
New feature-I created a password generator API Plugin. This generates a password based on a given length and excludes numbers and special characters if requested.

YousafZahid1:
I've submitted my own edition of password generator

YousafZahid1:
import random
import array

# maximum length of password needed
# this can be changed to suit your password length
MAX_LEN = 12

# declare arrays of the character that we need in out password
# Represented as chars to enable easy string concatenation
DIGITS = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
LOCASE_CHARACTERS = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h',
					'i', 'j', 'k', 'm', 'n', 'o', 'p', 'q',
					'r', 's', 't', 'u', 'v', 'w', 'x', 'y',
					'z']

UPCASE_CHARACTERS = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H',
					'I', 'J', 'K', 'M', 'N', 'O', 'P', 'Q',
					'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y',
					'Z']

SYMBOLS = ['@', '#', '$', '%', '=', ':', '?', '.', '/', '|', '~', '>',
		'*', '(', ')', '<']

# combines all the character arrays above to form one array
COMBINED_LIST = DIGITS + UPCASE_CHARACTERS + LOCASE_CHARACTERS + SYMBOLS

# randomly select at least one character from each character set above
rand_digit = random.choice(DIGITS)
rand_upper = random.choice(UPCASE_CHARACTERS)
rand_lower = random.choice(LOCASE_CHARACTERS)
rand_symbol = random.choice(SYMBOLS)

# combine the character randomly selected above
# at this stage, the password contains only 4 characters but
# we want a 12-character password
temp_pass = rand_digit + rand_upper + rand_lower + rand_symbol


# now that we are sure we have at least one character from each
# set of characters, we fill the rest of
# the password length by selecting randomly from the combined
# list of character above.
for x in range(MAX_LEN - 4):
	temp_pass = temp_pass + random.choice(COMBINED_LIST)

	# convert temporary password into array and shuffle to
	# prevent it from having a consistent pattern
	# where the beginning of the password is predictable
	temp_pass_list = array.array('u', temp_pass)
	random.shuffle(temp_pass_list)

[...]

PR #1111, opened 2023-05-03
Password Generator Created

Pasindu3683:
Created a password generator API plugin. Takes in input on length and Y/N for numbers and special character inclusion.
This addresses issue #1109

### DIFF
--- a/jarviscli/plugins/password_generator.py
+++ b/jarviscli/plugins/password_generator.py
@@ -1,0 +1,25 @@
+from plugin import plugin
+from colorama import Fore
+import requests
+
+@plugin("generate_password")
+
+def generate_password(jarvis, s):
+    length = int(jarvis.input("Desire password length: "))
+    inp = str(jarvis.input("Would you like to exclude numbers? (Y/N) : "))
+    if(inp == 'Y'):
+        exclude_numbers = True
+    else:
+        exclude_numbers = False
+
+    inp = bool(jarvis.input("Would you like to exclude special characters? (Y/N) : "))
+    if(inp == 'Y'):
+        exclude_special = True
+    else:
+        exclude_special = False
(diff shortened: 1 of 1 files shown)


jarviscli/plugins/password_generator.py
```diff
@@ -0,0 +1,25 @@
+from plugin import plugin
+from colorama import Fore
+import requests
+
+@plugin("generate_password")
+
```
ceramey1997, line 6: remove whitespace

jarviscli/plugins/password_generator.py
```diff
@@ -0,0 +1,25 @@
...
+def generate_password(jarvis, s):
+    length = int(jarvis.input("Desire password length: "))
+    inp = str(jarvis.input("Would you like to exclude numbers? (Y/N) : "))
+    if(inp == 'Y'):
+        exclude_numbers = True
+    else:
+        exclude_numbers = False
+
+    inp = bool(jarvis.input("Would you like to exclude special characters? (Y/N) : "))
```
ceramey1997, line 15: I'm fairly certain this would always evaluate to true as any string is a truthy value.

jarviscli/plugins/password_generator.py
```diff
@@ -0,0 +1,25 @@
+from plugin import plugin
+from colorama import Fore
+import requests
+
+@plugin("generate_password")
+
+def generate_password(jarvis, s):
+    length = int(jarvis.input("Desire password length: "))
+    inp = str(jarvis.input("Would you like to exclude numbers? (Y/N) : "))
```
ceramey1997, line 9: casting to a str here is un-necessary,  by default an input return type is str.

ceramey1997 (changes requested):

PR #1120, opened 2023-06-26
Fix #1116 evaluator plugin plotting

### LINKED ISSUE #1116: Unexpected output in Plot
i tried to plot y=constant it works well but when i do same with x it should make a vertical line instead it makes this

image

kapoorkrish:
Matplotlib was added to installer for improved plot rendering. Issue with graphing vertical lines (x=const) was fixed by using implicit plotting from sympy.
x=3

During debugging, it was discovered that equations such as x=y^2 led to multiple results that had to be plotted separately. Previously, only the first element in the results list was being returned, so the resulting plot was incomplete. Now, all plots returned in results will be plotted.
x=y^2

### DIFF
--- a/installer/requirements.txt
+++ b/installer/requirements.txt
@@ -10,6 +10,7 @@ geopy
 gitpython
 googletrans==3.1.0a0
 IMDbPY
+matplotlib
 mock
 nltk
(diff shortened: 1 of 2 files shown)


pnhofmann:
Thanks!

Issue #1129, opened 2023-08-26
fix: Ask user for column input again in connect_four instead of throwing error


daredevil0905:
## Description
In the `connect_four` plug-in, if the user enters a non-numeric value as input for `column`, then an error is thrown to the user which interrupts the gameplay and stops it. Instead of throwing `ValueError` to the user, we can simply ask them again for the column input till they enter a numeric input.

## Before
![Screenshot 2023-08-26 200414](https://github.com/sukeesh/Jarvis/assets/130159643/9820186b-2ffa-42ca-b458-eef6d26d3449)

daredevil0905:
Here is the PR for the fix as stated above

daredevil0905:
@pnhofmann is this fine or should I change something here?

PR #1130, opened 2023-08-26
fix: ask user again for valid column input in connect_four instead of throwing error

daredevil0905:
FIX: #1129 , #1131

Added the `ValueError` throwing code in a `while` loop in a `try-except` block till the user enters a numeric input.

## After fix
Screenshot 2023-08-26 200509


pnhofmann:
Looks good

PR #1133, opened 2023-09-10
new factorization plugin

raccoote:
Simple factorization plugin that takes an integer input, factors it into its prime factors, and communicates the result back using the jarvis.say() function. In case the user enters something that cannot be converted to an integer (e.g., a non-integer or negative number), it catches the ValueError and communicates an error message using jarvis.say().


pnhofmann:
Works good, thank you.
Thx

(Sorry for dealy)

Issue #1105, opened 2023-05-03
Suggestion to improve on calculator GUI


dinanz:
![image](https://user-images.githubusercontent.com/60189142/235822112-ca73eac0-7947-42af-8aed-7fc1ff103f42.png)
Suggestion to improve on the UI for the calculator.

cruncherino:
I can work on this

cruncherino:
Done, pull request #1135

pnhofmann:
#1135

PR #1135, opened 2023-09-28
Update User Interface of Calculator GUI

cruncherino:
This commit intends to update the user interface of the calculator.py plugin in Jarvis through changing the hex values and font families.

It seeks to address issue #1105 posted by dinanz on May 2.


pnhofmann:
Thanks

Issue #1138, opened 2023-10-12
New fruit information plugin


Sam0944:
A fruit plug-in which returns the information of fruit. Input a name of fruit and returns the nutritional composition table.

pnhofmann:
Sure!

WilliamYPY:
Hi! I can work on this plugin.

harshita280904:
Can I please be assigned to work on this

WilliamYPY:
> Sure!

Hi there, can you please assign this issue to me

PR #1145, opened 2023-10-26
implementation of new fruit plugin

harshita280904:
I added a new Fruit Plugin to the project. This would allow users to receive nutritional information about fruits. Additionally it would also let users to compare between two fruits in terms of their nutritional information. 
A new python file was added here.  jarviscli/plugins/nutrition_fruit
This plugin will be activated if the user calls either 'nutrition of fruits' or 'info on fruits'
the functionality was also tested running the jarvis.


harshita280904:
This addresses the issue #1138

pnhofmann:
Thanks

PR #1153, opened 2023-12-06
Add Hash Plugin for Jarvis

aandrewchung:
This pull request introduces a new hash plugin to the Jarvis codebase. The plugin enables users to hash strings and files using various hash functions like MD5, SHA1, and SHA256.

Key Changes:
- Added `hash.py` under `jarviscli/plugins`, implementing the hashing functionality.
- The plugin supports hashing strings directly from the user input or hashing contents of a specified file.
- Users can select from a range of hash algorithms for their specific needs.

I believe this addition will enhance the functionality of Jarvis by providing a useful utility for security and data verification purposes.

### DIFF
--- a/jarviscli/plugins/hash.py
+++ b/jarviscli/plugins/hash.py
@@ -1,0 +1,60 @@
+from plugin import plugin, complete
+import hashlib
+import os
+
+
+@complete("hash")
+@plugin("hash")
+def hash_data(jarvis, s: str) -> None:
+    """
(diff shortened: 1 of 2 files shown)


liu-julia:
Developed in collaboration with Andrew Chung (individual contributions removed in squash). I worked on test_hash.py and Andrew worked on hash.py

pnhofmann:
Nice work, thanks! Thanks for including tests ;)

Issue #1172, opened 2024-04-08
Visual enhancement for roulette game - spinning animation


lipinsks:
Using halo, I added visually pleasant "spinning animation", so the result of the roulette spin doesn't show up straight away. 
It's not a big feature, so I've already made it

PR #1173, opened 2024-04-08
Added spinning animation to roulette.py for visual enhancement

lipinsks:
Closes https://github.com/sukeesh/Jarvis/issues/1172

Using halo, I added visually pleasant "spinning animation", so the result of the roulette spin doesn't show up straight away.
Not a big feature, just a first contribution.


pnhofmann:
Nice! Thank you ;)

PR #1176, opened 2024-04-22
Update optional.py

anon15683:
Removed s at the end in Fasttext "instruction" of optional.py to stop exception while installing fasttext.

### DIFF
--- a/installer/optional.py
+++ b/installer/optional.py
@@ -57,7 +57,7 @@ Fasttext = {
     "name": "Fasttext language recognition",
     "pip": ['fasttext'],
     "description": "Fasttext is a text classification library capable of detecting 176 languages.",
-    "instructions": "https://github.com/facebookresearch/fastText/#requirements"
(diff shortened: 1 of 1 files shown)


anon15683:
Also, not sure how to do this, but for me all optional installs failed, the audio program is only called pyaudio or python3-pyaudio, but after installing it still won't see that it is installed. Same for fasttext, can not just be installed with pip and is not detected when installed.

pnhofmann:
Thanks!

> python3-pyaudio

Hmm, it needs to be installed to the virtualenv. Usually if you install the lib globally, it also installs all that is necessary to install in virtualenv.

For now I'll recommend to just ignore  and don't use speech recognition - right now it isn't good anyway. I'll plan to rework this whole part when I have time...

anon15683:
Haha, thanks for the quick reply. I will also see what I can do. My knowledge is technically a bit more than basic but not necessarily in this area. For now, I might also try it on windows and see what I can do

Issue #1178, opened 2024-05-18
Update README documentation with Jarvis capability list


shabazzj:
I think the README could benefit from having a short list (approximately 20 items) describing some of things that Jarvis can do in depth. Currently there is the outdated YouTube video and a short paragraph that gives a very general idea of some of the tasks it can perform for the user. I think this list would be beneficial because, at a glance, users can see if this is a program that they want to install and use. For developers, it would give them an idea of what plugins have already been installed, so that they may focus their skills on a different plugin idea. 

I saw that there was a new issue created to potentially update the YouTube video from 7 years ago. While I think this is also beneficial, I think the list may be better because it will be much easier to update the list as opposed to trying to make a new video every time a new feature/plugin gets added. Below is a picture showing what the updated README could look like.
![Jarvis_README_update_draft](https://github.com/sukeesh/Jarvis/assets/114114211/0aee1fc9-41e8-4a3d-8098-33b7a6e83192)

shabazzj:
I will assign this to myself if this is an acceptable issue to address.

PR #1181, opened 2024-06-06
Implement Issue #1178 -- Update README documentation

shabazzj:
This is an implementation for Issue #1178. This will update the README.md file to be more specify what tasks Jarvis can perform. This will allow potential users of the Jarvis app to know if it will be helpful for them. This will also let potential contributors know some of the major plugins that have been incorporated already.


pnhofmann:
Thanks!

PR #1191, opened 2024-08-04
Modified the setup order

MonnXI:
Modified the setup order so it takes python3 by default and after python which uses python2


pnhofmann:
Thanks!

PR #1194, opened 2024-09-02
Update caesar_cipher.py

KraVLonE:
- Update caesar shift values for simplicity and correctness
- Fix potential conversion error for non-ASCII characters
- Refactor Redundant code
- Add inline comments and function descriptions


pnhofmann:
Thanks!

Issue #1209, opened 2024-10-14
pylint CmdInterpreter.py file


leoscarlato:
I ran the pylint command in the CmdInterpreter.py file, and it indicated a lot of possible changes that could make the code even better, as shown below: 

pylint .\CmdInterpreter.py
************* Module jarviscli.CmdInterpreter
CmdInterpreter.py:323:0: C0325: Unnecessary parens after 'if' keyword (superfluous-parens)
CmdInterpreter.py:361:0: C0325: Unnecessary parens after 'if' keyword (superfluous-parens)
CmdInterpreter.py:379:0: C0301: Line too long (111/100) (line-too-long)
CmdInterpreter.py:477:0: C0325: Unnecessary parens after 'if' keyword (superfluous-parens)
CmdInterpreter.py:481:0: C0325: Unnecessary parens after 'if' keyword (superfluous-parens)
CmdInterpreter.py:485:0: C0325: Unnecessary parens after 'if' keyword (superfluous-parens)
CmdInterpreter.py:1:0: C0114: Missing module docstring (missing-module-docstring)
CmdInterpreter.py:1:0: C0103: Module name "CmdInterpreter" doesn't conform to snake_case naming style (invalid-name)
CmdInterpreter.py:10:0: E0401: Unable to import 'packages.memory.memory' (import-error)
CmdInterpreter.py:11:0: E0401: Unable to import 'PluginManager' (import-error)
CmdInterpreter.py:12:0: E0401: Unable to import 'utilities' (import-error)
CmdInterpreter.py:13:0: E0401: Unable to import 'utilities.animations' (import-error)
CmdInterpreter.py:14:0: E0401: Unable to import 'utilities.GeneralUtilities' (import-error)
CmdInterpreter.py:15:0: E0401: Unable to import 'utilities.notification' (import-error)
CmdInterpreter.py:16:0: E0401: Unable to import 'utilities.voice' (import-error)
CmdInterpreter.py:19:0: R0205: Class 'JarvisAPI' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance) 
CmdInterpreter.py:62:4: R0913: Too many arguments (6/5) (too-many-arguments)
CmdInterpreter.py:62:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments)
CmdInterpreter.py:78:29: C0209: Formatting a regular string which could be an f-string (consider-using-f-string)
[...]

PR #1210, opened 2024-10-15
pylint suggestions - CmdInterpreter.py

leoscarlato:
As I've said in the issue #1209, I ran the pylint command in the `CmdInterpreter.py` file, and it indicated a lot of possible changes that could make the code even better.

Before the changes I`ve made, the overall code rating was at **6.73/10**.

After some changes that I've made, I was able to implement some of the suggestions that the tool suggested, therefore making the code rating even better, reaching a value of **7.77/10**.

image

I wasn't able to implement some suggestions that are also indicated in the picture, but maybe someone else could continue this task in the future.


pnhofmann:
Thanks

PR #1211, opened 2024-10-16
Update numbers_api.py

Jakub-Haus:
new functionality added:
numbersapi random trivia
numbersapi random date
numbersapi random year
numbersapi random number

numbersapi 1 2 3 still works like before


pnhofmann:
thanks!

Issue #1214, opened 2024-10-18
NEW FEATURE: Add a Color Converter Plugin for RGB, HEX, and HSL formats


nicolasgms:
I would like to propose a new feature for a Color Converter plugin that allows users to convert between RGB, HEX, and HSL color formats. This feature will help users easily switch between these common color representations, often needed in web development, design, and other technical fields.

Details of the plugin:
Convert from HEX to RGB and HSL
Convert from RGB to HEX and HSL
Convert from HSL to RGB and HEX
Validation for color inputs and proper error handling

Usage:
Users choose the original format (RGB, HEX, HSL).
Users select the target format to convert to.
The plugin outputs the converted color value.

I am interested in contributing this feature. Please let me know if it's okay.

pnhofmann:
Nice!

PR #1216, opened 2024-11-04
Feat: color converter

nicolasgms:
#1214  This PR introduces a new plugin for Jarvis that provides color code conversion functionalities. The plugin allows users to convert color values between HEX, RGB, and HSL formats quickly. This feature is particularly useful for developers and designers working with color codes, as it provides an easy way to switch between formats without leaving the command line.


pnhofmann:
Thanks!

PR #1223, opened 2024-12-06
Add Geolocation plugin

ip2location:
This plugin will helps to geolocate an IP address without an API key, by just entering an IP address.

A quick example would be:

```
geolocation 8.8.8.8
Lookup 8.8.8.8 now...
It's located at Mountain View, California, United States of America.
It's coordinates are (37.38605, -122.08385).
It's owned by Google LLC with the ASN 15169.
```

### DIFF
--- a/jarviscli/plugins/geolocation.py
+++ b/jarviscli/plugins/geolocation.py
@@ -1,0 +1,15 @@
+import json
+import requests
+
+@require(network=True)
+@plugin('geolocation')
+def geolocation(jarvis, s):
+    """ Helps to geolocate an IP address """
+    base_url = "https://api.ip2location.io/"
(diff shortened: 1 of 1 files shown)


pnhofmann:
Hey!

this is an interesting api!
Out of curiosity, what is the difference between ip2location.com (from your bio) and ip2location.io (from your source code) - because appears to be different sites with different logos.

Additionally, what is about attribution:
https://www.ip2location.io/pricing#attribution

ip2location:
@pnhofmann 

IP2Location.com is database subscription and IP2Location.io is for API. Both owned by the same company.

pnhofmann:
Thanks!

PR #1224, opened 2024-12-07
feature implemented: markdown to pdf

khunya22:
Added a feature for converting markdown to pdf. 

Command: "markdown to pdf"

The full path of the file is required.


pnhofmann:
Thanks!

Issue #1189, opened 2024-07-27
JarvisCLI website and discord


smitthakore:
I guess having a website will attract more non dev users, allowing them to directly download jarviscli. Also a discord for contributors to communicate or express their ideas and for non dev users to share their bugs and report them is utmost needed.

aej6033:
Could I be assigned to this issue of developing the website? 

I am CS student from UMICH looking to work on issues here for one of my upper level courses.

PR #1228, opened 2024-12-09
Jarvis Website

aej6033:
Response to issue: https://github.com/sukeesh/Jarvis/issues/1189

I've created a website for Jarvis for people to download Jarvis from and a question form.
It's built on python and django.

Let me know if there are any questions.

Dynamic download. When the user selects to download Jarvis the deliverable is built then and served to the user.


pnhofmann:
Good work!
Sorry for really late reply. And to be honest, I don't really know what to do with your PR yet - since actually I don't want to host a Jarvis website.
But - as an idea - would mabye a nice base for interacting with Jarvis using the django backend? Unfortunately don't know, how much time I'll have working on this.

realblobii (approved):

Issue #1232, opened 2025-01-06
Bug : AttributeError: 'coroutine' object has no attribute 'src'


git-akispa:
# **Error message:**

`AttributeError: 'coroutine' object has no attribute 'src'
C:\Users\MyUser\Desktop\Jarvis\jarviscli\CmdInterpreter.py:297: RuntimeWarning: coroutine 'Translator.translate' was never awaited
  print(Fore.RESET)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback`

# **How to reproduce:**

    1. Clone from the main branch.
    2. Run the command python installer
    3. Run the command ./jarvis
    4. Run any translate request
 
# **Expected behaviour:**

No Error

# **Machine details:**

Windows 11 Pro, Running commands from CMD within the Jarvis installation directory.

# **Additional details:**

Googletrans was recently updated to version 4.0.2, and as the link below shows, the way you use the module has changed. 
https://pypi.org/project/googletrans/

# **Proposed fix:**

I have made the changes and making a Pull Request changing the code, while also implementing some new features:

    1. Adding the word auto as a source language for automatic language detection
    2. Adding the ability to add a path as text input for .txt file translation*
    3. Adding the ability to save the output to a file**
    
    *The maximum number of characters the API can handle in one request is 15k, so I am implementing the ability to break the text 
    into smaller chunks and translate them separately. The separation is done at a new line character for minimum semantic loss.
    
    **The file is at a default path under the Jarvis directory and also for the case where the input is a file with more than 10k characters 
    the output is stored in the file instead of being printed in the console as it would clutter it.

pnhofmann:
#1233

PR #1233, opened 2025-01-06
Update translation.py to work with the newer googletrans module

git-akispa:
Fixing a bug in the `translation` plugin where it is not working with the new default version of googletrans as mentioned in #1232.

Also implementing some new features :

1. Adding the word auto as a source language for automatic language detection
2. Adding the ability to add a path as text input for .txt file translation*
3. Adding the ability to save the output to a file**

*The maximum number of characters the API can handle in one request is 15k, so I am implementing the ability to break the text  into smaller chunks and translate them separately. The separation is done at a new line character for minimum semantic loss.

**The file is at a default path under the Jarvis directory and also for the case where the input is a file with more than 10k characters  the output is stored in the file instead of being printed in the console as it would clutter it.

### DIFF
--- a/jarviscli/plugins/translate.py
+++ b/jarviscli/plugins/translate.py
@@ -2,6 +2,8 @@ from plugin import plugin, require, alias
 from googletrans import Translator
 from googletrans.constants import LANGCODES, LANGUAGES, SPECIAL_CASES
 import nltk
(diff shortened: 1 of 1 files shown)


pnhofmann:
Thanks!

Issue #1235, opened 2025-01-19
python installer does not work properly


ArifODCL:
============== Preparing virtualenv ==============
________________________________________
Shell: python3 --version
Python 3.12.3

Shell: Exit OK
----------------------------------------

________________________________________
Shell: virtualenv --python=python3 --version
virtualenv 20.25.0+ds from /usr/lib/python3/dist-packages/virtualenv/__init__.py

Shell: Exit OK
----------------------------------------

________________________________________
Shell: env/bin/python --version
Python 3.12.3

Shell: Exit OK
----------------------------------------

============ Installing requirements =============
________________________________________
Shell: env/bin/pip install -U -r installer/requirements.txt
Ignoring pathlib2: markers 'python_version < "3.4"' don't match your environment
Ignoring psutil: markers 'sys_platform == "win32"' don't match your environment
Ignoring pync: markers 'sys_platform == "darwin"' don't match your environment
Ignoring pypiwin32: markers 'sys_platform == "win32"' don't match your environment
Ignoring pyreadline: markers 'sys_platform == "win32"' don't match your environment
Ignoring pywin32: markers 'sys_platform == "win32"' don't match your environment
Ignoring win10toast: markers 'sys_platform == "win32"' don't match your environment
Ignoring windows-curses: markers 'sys_platform == "win32"' don't match your environment
Collecting archey4==4.6.0.post1 (from -r installer/requirements.txt (line 1))
  Using cached archey4-4.6.0.post1-py3-none-any.whl.metadata (1.0 kB)
Collecting beautifulsoup4 (from -r installer/requirements.txt (line 2))
  Using cached beautifulsoup4-4.12.3-py3-none-any.whl.metadata (3.8 kB)
Collecting climage (from -r installer/requirements.txt (line 3))
  Using cached climage-0.2.2-py3-none-any.whl.metadata (8.2 kB)
Collecting colorama (from -r installer/requirements.txt (line 4))
  Using cached colorama-0.4.6-py2.py3-none-any.whl.metadata (17 kB)
Collecting distro (from -r installer/requirements.txt (line 5))
  Using cached distro-1.9.0-py3-none-any.whl.metadata (6.8 kB)
Collecting flake8 (from -r installer/requirements.txt (line 6))
  Using cached flake8-7.1.1-py2.py3-none-any.whl.metadata (3.8 kB)
Collecting forex-python (from -r installer/requirements.txt (line 7))
[...]

PR #1240, opened 2025-02-03
Issue #1235: Fix playsound installation on Python 3.12+

jonasyr:
- Updates installer/requirements.txt to install a forked version of playsound when running Python 3.12 or newer.
- Fixes issue #1235 where the default playsound package fails to build on Python 3.12+.
- Uses environment markers to ensure compatibility with older Python versions.


pnhofmann:
thanks

PR #1241, opened 2025-02-16
Creation of top_music.py plugin

SethSainz:
Add topmusic plugin using Spotify API for top 10 songs by genre. Has a pool of 28 different genres to choose from. Additionally, added spotipy to the requirements.txt file so that the plugin I created would be supported and recognized by Jarvis.

Using a command, such as "topmusic rock" the user will be provided by Jarvis with a list of the current top 10 songs of the provided genre on Spotify.

### DIFF
--- a/jarviscli/plugins/top_music.py
+++ b/jarviscli/plugins/top_music.py
@@ -1,0 +1,87 @@
+from plugin import plugin, require
+import spotipy
+from spotipy.oauth2 import SpotifyClientCredentials
+
+# Set up your credentials
+client_id = '9432469329f94c449bc36b3273f7285a'
+client_secret = '35d017ceebeb4eea83cd74ff5ee091ac'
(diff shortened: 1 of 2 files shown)


jarviscli/plugins/top_music.py
```diff
@@ -0,0 +1,87 @@
+from plugin import plugin, require
+import spotipy
+from spotipy.oauth2 import SpotifyClientCredentials
+
+# Set up your credentials
+client_id = '9432469329f94c449bc36b3273f7285a'
+client_secret = '35d017ceebeb4eea83cd74ff5ee091ac'
```
pnhofmann, line 7: Please don't put this in code. Please generate a new secret + make this secret invalid. We don't host credentials - features that need api key are allowed, but the user must provide an own api key.

PR #1245, opened 2025-02-17
Word Chain Game Plugin

mittvansh:
Created a new plugin game that uses Dictionary API to play a vocabulary building word chain game that can be played with Jarvis and it keeps score.


pnhofmann:
Nice, thanks ;)

PR #1248, opened 2025-02-21
 added tests for parse_date

AlessandroColi:
added tests for working cases in parse_date to provide a almost 90% cobverage (24 of the 26 branches before the refactor)

### DIFF
--- a/jarviscli/tests/test_parse_date.py
+++ b/jarviscli/tests/test_parse_date.py
@@ -1,0 +1,108 @@
+import unittest
+from datetime import datetime, timedelta
+from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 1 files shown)


mmori7 (commented):
Documentation: Consider adding more inline comments, especially in the helper method assert_datetime_equal, to explain the logic for comparing datetime objects. This will make it easier for others to understand the intent.

Additional Test Cases: While the current test cases are exhaustive, you might want to include edge cases, such as:

Invalid date and time formats to check how the parse_date function handles errors.
Cases with ambiguous inputs (e.g., "1/2/03", which could imply different formats based on locale).

pnhofmann:
Thanks!

PR #1249, opened 2025-02-21
Refactor date parser

AlessandroColi:
Replaced the multiple if statements with a dictionary of regex patterns for formats and a map for time units to their respective handlers

contains also same tests as PR #1248


AlessandroColi:
> Test ambiguous inputs, such as "05/06/2025" (could mean MM/DD/YYYY or DD/MM/YYYY depending on locale).

True but that's how it was in the original code, so I kept it that way, so I belive to have not removed any functionality.

Ronitsabhaya75:
I see what you mean

Issue #1261, opened 2025-08-07
Issue occurred at Personality test


gadalova:
Your personality is: INFJ
Redirecting to your personality analysis                 in 3s...

**Some error occurred, please open an issue on github!**
Here is error:

Traceback (most recent call last):
  File "C:\Users\gadal\Jarvis\jarviscli\CmdInterpreter.py", line 285, in try_do
    do(self, s)
    ~~^^^^^^^^^
  File "C:\Users\gadal\Jarvis\jarviscli\plugin.py", line 240, in run
    self._backend[0](jarvis.get_api(), s)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\gadal\Jarvis\jarviscli\plugins\personality.py", line 100, in __call__
    self.open_analysis()
    ~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\gadal\Jarvis\jarviscli\plugins\personality.py", line 65, in open_analysis
    webbrowser.open_new(url.format(self.type.lower()))
                                   ^^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'lower'

lrsierra:
Student Contributor: I’d like to work on this. Will reproduce the personality_test issue on macOS/Py3.9 and submit a minimal fix with tests. Please assign.

PR #1262, opened 2025-08-07
Fix AttributeError in personality plugin

gadalova:
To fix the following issue:

**Issue occurred at Personality test #1261**

Line 65 was changed from this:

webbrowser.open_new(url.format(self.type.lower()))

to this:

webbrowser.open_new(url.format(self.type[0].lower()))


pnhofmann:
Thanks

PR #1264, opened 2025-08-08
Add Windows support for camera plugin

jarvis905:
Added Windows support for the camera plugin to open the camera app using Jarvis


pnhofmann:
Thanks ;)

